PR #23, opened 2026-03-04
fix(windows): zip extraction fails when primary download server is down

evelaa123:
<html>
<body>
<html><head></head><body><h2>Problem</h2>
<p>On Windows, <code>ensureBinary()</code> consistently fails with <strong>“file in use by another process”</strong> when extracting the downloaded Chromium zip. This happens every time the primary download server (<code>cloakbrowser.dev</code>) is unavailable and the fallback to GitHub Releases is triggered.</p>
<p>The error appears as:</p>
<pre><code>The process cannot access the file
'C:\Users\...\.cloakbrowser\_download_xxx.zip'
because it is being used by another process.
</code></pre>
<p>This makes CloakBrowser <strong>completely unusable on first install</strong> for any Windows user when the primary CDN is down — which currently happens consistently (<code>terminated</code> error on every attempt).</p>
<h2>Root cause</h2>
<p>The bug is in <code>downloadFile()</code> inside <code>downloadAndExtract()</code>. Here is the exact sequence:</p>
<p><strong>Step 1</strong> — <code>downloadFile(primaryUrl, tmpPath)</code> is called:</p>
<ul>
<li><code>createWriteStream(tmpPath)</code> opens a file handle (handle #1)</li>
<li><code>fetch(primaryUrl)</code> starts but the server is down</li>
<li>The fetch throws <code>terminated</code></li>
<li>The function exits via the thrown error</li>
<li><strong><code>fileStream</code> is never closed or destroyed — handle #1 is leaked</strong></li>
</ul>
<p><strong>Step 2</strong> — <code>downloadFile(fallbackUrl, tmpPath)</code> is called on the same <code>tmpPath</code>:</p>
<ul>
<li><code>createWriteStream(tmpPath)</code> opens a NEW file handle (handle #2) on the same file</li>
<li><code>fetch(fallbackUrl)</code> succeeds, 526 MB downloaded</li>
<li><code>fileStream.end(() =&gt; resolve())</code> is called — handle #2 is closed</li>
<li>Download completes successfully</li>
</ul>
<p><strong>Step 3</strong> — <code>extractArchive(tmpPath, ...)</code> is called:</p>
<ul>
<li><code>extractZip()</code> spawns PowerShell to extract the zip</li>
<li>PowerShell tries to open the file but <strong>handle #1 from Step 1 is still open</strong></li>
<li>Windows returns <code>ERROR_SHARING_VIOLATION</code></li>
[...]

PR #29, opened 2026-03-05
fix(python): reduce download connect timeout to 10s, read to 60s for …

evelaa123:
## Problem

When the primary download server (cloakbrowser.dev) is unreachable, `httpx.stream()` waits the full `DOWNLOAD_TIMEOUT` (600 seconds / 10 minutes) before falling back to GitHub Releases. First-time users on any platform sit staring at a frozen terminal for 10+ minutes.

## Root cause

`DOWNLOAD_TIMEOUT = 600.0` is a single float, which httpx applies to **all** phases equally — connect, read, write, pool. A dead server that accepts the TCP connection but never sends data holds the stream open for the full 600s on the read phase.

## Fix

Replace the single float with granular timeouts:

```python
DOWNLOAD_TIMEOUT = httpx.Timeout(connect=10.0, read=60.0, write=10.0, pool=10.0)
```

- **connect=10s** — fail fast if server doesn't accept connection
- **read=60s** — if server connects but sends no data, fail after 60s (not 600s)
- **write/pool=10s** — reasonable defaults

Note: `read=60s` is conservative on purpose — slower connections downloading the 526 MB binary won't be interrupted since the read timer resets with each received chunk. This could be reduced further (e.g. `read=30s`) if needed.

## Before / After

| Scenario | Before | After |
|---|---|---|
| Primary down → fallback → complete | ~692s (11.5 min) | ~170s (2.8 min) |
| Primary works → complete | ~90s | ~90s (unchanged) |

Tested on Windows 11, Python 3.10, cloakbrowser 0.3.8.

## Note

The JS wrapper had a related issue (PR #23 — file handle leak causing extraction failure). The Python wrapper doesn't have the file handle bug (context manager `with open()` closes correctly), but the timeout makes first-run UX equally bad when the primary server is down.

### DIFF
--- a/cloakbrowser/download.py
+++ b/cloakbrowser/download.py
@@ -44,7 +44,7 @@ from .config import (
 logger = logging.getLogger("cloakbrowser")
 
 # Timeout for download (large binary, allow 10 min)
-DOWNLOAD_TIMEOUT = 600.0
+DOWNLOAD_TIMEOUT = httpx.Timeout(connect=10.0, read=60.0, write=10.0, pool=10.0)
 
 # Auto-update check interval (1 hour)
 UPDATE_CHECK_INTERVAL = 3600


Cloak-HQ:
Thanks for this — clean fix, well-documented PR. Merged.

PR #38, opened 2026-03-09
feat(proxy): add ProxyRotator with multi-strategy rotation, health tracking, and auto-failover

evelaa123:
## Summary

Adds `ProxyRotator` — a thread-safe proxy pool with automatic rotation, health tracking, and failover for both Python and JavaScript/TypeScript.

## Background

The design was inspired by [Scrapling's ProxyRotator](https://github.com/D4Vinci/Scrapling), which provides cyclic proxy rotation with custom strategy callbacks for its spider framework. CloakBrowser's implementation is written from scratch and differs significantly:

- **Built-in strategies** instead of callback-based: four strategies are built in (round-robin, random, least-used, least-failures) rather than requiring the user to write a strategy function.
- **Health tracking with automatic cooldown:** Scrapling's rotator is stateless — it just cycles through a list. Ours tracks per-proxy failure counts, consecutive failures, and automatically sidelines broken proxies for a configurable cooldown period, then recovers them.
- **Sticky sessions:** Scrapling has no concept of reusing the same proxy for N consecutive requests. Ours supports `sticky_count` to keep the same proxy for a batch of requests (useful for session-based websites).
- **Dynamic pool management:** proxies can be added/removed at runtime without restarting. Scrapling's list is static.
- **Thread safety:** all operations are protected by `threading.Lock` (Python). Scrapling's rotator is not thread-safe.
- **Credential masking:** `stats()` and all log output automatically mask usernames and passwords. Scrapling does not mask credentials.
- **SOCKS5 validation:** rejects `socks5://user:pass@host:port` at construction time with a clear error message, because Chromium does not support SOCKS5 authentication. This prevents a confusing launch-time crash.
- **Bare proxy format:** accepts `user:pass@host:port` without a scheme prefix, which is common in proxy provider dashboards.
- **Dual implementation:** both Python and TypeScript, matching CloakBrowser's architecture. Scrapling is Python-only.
- **Context manager / withSession:** auto-reports success on normal return, failure on exception. Scrapling relies on the spider framework's retry system instead.

In short: Scrapling's ProxyRotator was the starting inspiration for "a class that rotates proxies," but the actual implementation, API, and feature set are entirely different and written from scratch for CloakBrowser's use case (standalone browser automation, not a spider framework).

### Rotation Strategies

- **round-robin** (default): cycles through proxies in order, skipping any that are in cooldown
- **random**: picks a random available proxy each time
- **least-used**: picks the proxy with the fewest total uses (distributes load evenly)
- **least-failures**: picks the proxy with the fewest total failures (avoids unreliable proxies)

### Health Tracking

Each proxy tracks: `use_count`, `fail_count`, `consecutive_fails`, `last_used`, `last_failed`, `cooldown_until`.

When a proxy reaches `max_failures` consecutive failures (default: 3), it is automatically placed on cooldown for `cooldown` seconds (default: 300). During cooldown it is skipped by all strategies. When cooldown expires, the proxy becomes available again. Calling `report_success()` immediately resets the consecutive failure counter and removes any cooldown.

If all proxies are in cooldown simultaneously, `next()` raises `RuntimeError` with a clear message.

### Sticky Sessions

Set `sticky_count=N` to reuse the same proxy for N consecutive `next()` calls before rotating. Useful for websites that track sessions by IP. If the sticky proxy fails, rotation is forced immediately.

### Dynamic Pool

- `add(proxy)` — add a proxy to the pool at runtime
- `remove(proxy)` — remove a proxy (raises if not found or would empty the pool)
- Round-robin index is clamped after removal to prevent index errors
- Sticky state is cleared if the sticky proxy is removed

### Context Manager / withSession

Python:
```python
with rotator.session() as proxy:
    browser = launch(proxy=proxy)
    # if this block raises, report_failure() is called automatically
    # if it completes normally, report_success() is called
```

JavaScript:
```typescript
await rotator.withSession(async (proxy) => {
    const browser = await launch({ proxy });
    // same auto-reporting behavior
});
```

### Credential Masking

`stats()` output and all internal logging automatically masks credentials:
- `http://user:pass@host:8080` → `http://***:***@host:8080`
- `user:pass@host:8080` (bare format) → `***:***@host:8080`
- `socks5://host:8080` (no creds) → unchanged
- Dict key format `http://host:8080||username` → `http://host:8080||***`

### SOCKS5 Validation

Chromium does not support SOCKS5 proxy authentication. Rather than failing with a cryptic error at browser launch time, `ProxyRotator` validates proxies at construction and `add()` time:

- `socks5://host:port` — accepted (no auth)
- `socks5://user:pass@host:port` — raises `ValueError` immediately
- `{"server": "socks5://host:port", "username": "u", "password": "p"}` — raises `ValueError` immediately

### Integration with launch()

```python
from cloakbrowser import ProxyRotator, launch

rotator = ProxyRotator([
    "http://user:pass@proxy1:8080",
    "http://user:pass@proxy2:8080",
], strategy="round_robin")

browser = launch(proxy=rotator.next())
page = browser.new_page()
page.goto("https://example.com")
rotator.report_success(rotator.current())
browser.close()
```

The `current()` method returns the proxy that was last selected by `next()`, so you can report success/failure after using it.

### Python (`cloakbrowser.proxy_rotator`)

```python
class ProxyRotator:
    def __init__(self, proxies, strategy="round_robin", cooldown=300.0, max_failures=3, sticky_count=1)
    def next(self) -> str | dict
    def current(self) -> str | dict | None
    def report_success(self, proxy) -> None
    def report_failure(self, proxy) -> None
    def session(self) -> ContextManager[str | dict]
    def stats(self) -> list[dict]
    def reset(self) -> None
    def add(self, proxy) -> None
    def remove(self, proxy) -> None
    @property
    def available_count(self) -> int
```

### JavaScript/TypeScript (`cloakbrowser`)

```typescript
class ProxyRotator {
    constructor(proxies: ProxyValue[], options?: ProxyRotatorOptions)
    next(): ProxyValue
    current(): ProxyValue | null
    reportSuccess(proxy: ProxyValue): void
    reportFailure(proxy: ProxyValue): void
    withSession<T>(fn: (proxy: ProxyValue) => Promise<T>): Promise<T>
    stats(): ProxyStats[]
    reset(): void
    add(proxy: ProxyValue): void
    remove(proxy: ProxyValue): void
    get availableCount(): number
    get size(): number
    static proxyKey(proxy: ProxyValue): string
}
```

## Tests

| Suite | Count | Status |
|-------|-------|--------|
| Python unit (pytest) | 49 | ✅ |
| JS unit (vitest) | 45 | ✅ |
| Python real-proxy integration | 7 | ✅ |
| JS real-proxy integration | 9 | ✅ |
| **Total** | **110** | **✅** |

Tests cover: all four strategies, health tracking with cooldown expiration, sticky sessions, context manager success/failure paths, dynamic pool add/remove edge cases, credential masking for all proxy formats, SOCKS5 validation, thread safety (10 threads × 100 calls), bare proxy format handling, round-robin index clamping after removal, and real browser integration (headless Chromium verifying external IP via ipify.org).

### Added
- `cloakbrowser/proxy_rotator.py` — Python implementation
- `js/src/proxy-rotator.ts` — TypeScript implementation
- `tests/test_proxy_rotator.py` — Python unit tests (49 tests)
- `js/tests/proxy-rotator.test.ts` — JS unit tests (45 tests)
- `tests/test_proxy_real.py` — Python real-proxy integration tests
- `tests/test_proxy_real.mjs` — JS real-proxy integration tests
- `examples/proxy_rotation.py` — Python usage example
- `examples/proxy_verify.py` — Python proxy verification script
- `js/examples/proxy-rotation.ts` — TypeScript usage example

### Modified
- `cloakbrowser/__init__.py` — export ProxyRotator
- `cloakbrowser/browser.py` — integrate rotator with launch()
- `js/src/index.ts` — export ProxyRotator, maskProxy, resolveProxyRotator
- `js/src/proxy.ts` — integrate rotator resolution
- `js/src/types.ts` — add ProxyRotator-related types
- `js/src/playwright.ts` — resolve rotator in Playwright launcher
- `js/src/puppeteer.ts` — resolve rotator in Puppeteer launcher
- `tests/test_proxy.py` — add bare proxy format parsing tests

### DIFF
--- a/cloakbrowser/__init__.py
+++ b/cloakbrowser/__init__.py
@@ -14,6 +14,7 @@ Usage:
 from .browser import launch, launch_async, launch_context, launch_persistent_context, launch_persistent_context_async, ProxySettings
 from .config import CHROMIUM_VERSION, get_default_stealth_args
 from .download import binary_info, check_for_update, clear_cache, ensure_binary
+from .proxy_rotator import ProxyRotator, Strategy as ProxyStrategy
 from ._version import __version__
 
 # Human-like behavioral layer (optional)
@@ -41,6 +42,8 @@ __all__ = [
     "CHROMIUM_VERSION",
     "get_default_stealth_args",
     "ProxySettings",
+    "ProxyRotator",
+    "ProxyStrategy",
     "HumanConfig",
     "resolve_human_config",
     "__version__",
--- a/cloakbrowser/browser.py
+++ b/cloakbrowser/browser.py
@@ -22,6 +22,7 @@ from urllib.parse import unquote, urlparse, urlunparse
 
 from .config import DEFAULT_VIEWPORT, get_default_stealth_args
 from .download import ensure_binary
+from .proxy_rotator import ProxyRotator
 
 logger = logging.getLogger("cloakbrowser")
 
@@ -51,7 +52,7 @@ class ProxySettings(_ProxySettingsRequired, total=False):
 
 def launch(
     headless: bool = True,
-    proxy: str | ProxySettings | None = None,
+    proxy: str | ProxySettings | ProxyRotator | None = None,
     args: list[str] | None = None,
     stealth_args: bool = True,
     timezone: str | None = None,
@@ -102,6 +103,7 @@ def launch(
     """
     sync_playwright = _import_sync_playwright(_resolve_backend(backend))
 
+    proxy = _resolve_proxy_rotator(proxy)
     binary_path = ensure_binary()
     timezone, locale = _maybe_resolve_geoip(geoip, proxy, timezone, locale)
     chrome_args = _build_args(stealth_args, args, timezone=timezone, locale=locale)
@@ -139,7 +141,7 @@ def launch(
 
 async def launch_async(  # noqa: C901
     headless: bool = True,
-    proxy: str | ProxySettings | None = None,
+    proxy: str | ProxySettings | ProxyRotator | None = None,
     args: list[str] | None = None,
     stealth_args: bool = True,
     timezone: str | None = None,
@@ -185,6 +187,7 @@ async def launch_async(  # noqa: C901
     """
     async_playwright = _import_async_playwright(_resolve_backend(backend))
 
+    proxy = _resolve_proxy_rotator(proxy)
     binary_path = ensure_binary()
     timezone, locale = _maybe_resolve_geoip(geoip, proxy, timezone, locale)
     chrome_args = _build_args(stealth_args, args, timezone=timezone, locale=locale)
@@ -223,7 +226,7 @@ async def launch_async(  # noqa: C901
 def launch_persistent_context(
     user_data_dir: str | os.PathLike,
     headless: bool = True,
-    proxy: str | ProxySettings | None = None,
+    proxy: str | ProxySettings | ProxyRotator | None = None,
     args: list[str] | None = None,
     stealth_args: bool = True,
     user_agent: str | None = None,
@@ -279,6 +282,7 @@ def launch_persistent_context(
     """
     sync_playwright = _import_sync_playwright(_resolve_backend(backend))
 
+    proxy = _resolve_proxy_rotator(proxy)
     timezone = _migrate_timezone_id(timezone, kwargs)
 
     binary_path = ensure_binary()
@@ -336,7 +340,7 @@ def launch_persistent_context(
 async def launch_persistent_context_async(
     user_data_dir: str | os.PathLike,
     headless: bool = True,
-    proxy: str | ProxySettings | None = None,
+    proxy: str | ProxySettings | ProxyRotator | None = None,
     args: list[str] | None = None,
     stealth_args: bool = True,
     user_agent: str | None = None,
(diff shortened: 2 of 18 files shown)


Cloak-HQ:
Hey, thanks for the detailed PR and the solid test coverage — this is clearly well thought out.

After reviewing, we're going to pass on merging this one. The core reason: proxy rotation is orchestration logic that lives *between* browser sessions, not inside the launcher. Since Chrome binds the proxy at process start (it's a `--proxy-server` flag — there's no mid-session swap), the rotator's job is purely deciding what to pass into the next `launch()` call. That's the caller's responsibility, not the wrapper's.

Compare this to `humanize`, which patches browser behavior *during* a session and can't exist outside of it. The rotator can — it would work identically as a standalone utility or even just a few lines in user code.

We want to keep the wrapper focused on what only it can do: launching a stealthy browser. For multi-session workflows, check out [CloakBrowser Manager](https://github.com/CloakHQ/CloakBrowser-Manager) — self-hosted profile management with persistent sessions, proxies, and noVNC.

One thing we do want to take from this PR: the bare proxy format fix (`user:pass@host:port` without a scheme). We've already landed that separately. Thanks for surfacing it.

Willbelyeu:
I'm curious, what is the main use case for this? It is an interesting idea but I don't see where it would be used specifically

PR #60, opened 2026-03-15
fix: ensure pw.stop() runs even if browser.close() raises or is cancelled

dgtlmoon:
Greetings from https://github.com/dgtlmoon/changedetection.io <3

**Problem - Session cleanups can be improved**

All four launch functions patch browser.close() / context.close() to also call pw.stop(), which is the right approach. However the patched functions have no error handling:

```
  async def _close_with_cleanup() -> None:
      await _original_close()   # if this raises or is CancelledError...
      await pw.stop()           # ...this line is never reached
```

This means the Playwright Node.js subprocess can be left running if:

- The Chromium process hangs and the caller uses asyncio.wait_for(browser.close(), timeout=N)  the timeout cancels the coroutine with CancelledError before pw.stop() runs
- _original_close() raises any exception (renderer crash, already-closed browser, etc.)

The result is an orphaned playwright Node process, open DevTools sockets, and /tmp/playwright-* profile directories that are only cleaned up when that process eventually dies on its own.

**Fix**

Wrap each patched cleanup function in try/finally so pw.stop() is guaranteed to run regardless of what _original_close() does:

```
  # Before
  async def _close_with_cleanup() -> None:
      await _original_close()
      await pw.stop()

  # After
  async def _close_with_cleanup() -> None:
      try:
          await _original_close()
      finally:
          await pw.stop()
```

Same one-line change applies to the sync variants (launch, launch_persistent_context, launch_persistent_context_async, launch_context).

Why this matters in practice

Callers that defensively wrap browser.close() with a timeout — which is the correct thing to do for a hung page — currently get a resource leak as a side-effect of doing the right thing:

```
  # Caller does the right thing...
  try:
      await asyncio.wait_for(browser.close(), timeout=5.0)
  except asyncio.TimeoutError:
      pass
  # ...but pw.stop() was never called, playwright process stays alive
```
After this fix, pw.stop() runs in all cases.

### DIFF
--- a/cloakbrowser/browser.py
+++ b/cloakbrowser/browser.py
@@ -120,8 +120,10 @@ def launch(
     _original_close = browser.close
 
     def _close_with_cleanup() -> None:
-        _original_close()
-        pw.stop()
+        try:
+            _original_close()
+        finally:
+            pw.stop()
 
     browser.close = _close_with_cleanup
 
@@ -203,8 +205,10 @@ async def launch_async(  # noqa: C901
     _original_close = browser.close
 
     async def _close_with_cleanup() -> None:
-        await _original_close()
-        await pw.stop()
+        try:
+            await _original_close()
+        finally:
+            await pw.stop()
 
     browser.close = _close_with_cleanup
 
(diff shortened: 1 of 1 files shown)


Cloak-HQ:
Thanks @dgtlmoon, clean fix. Merged.

Nice to see changedetection.io using CloakBrowser!

dgtlmoon:
Anytime :) btw, looks like `akamaighost` still rely on IP-reputation and/or proxy detection and/or VPN detection I think, nice to see some basic sites working again with cloak :) try https://www.homedepot.com/b/Heating-Venting-Cooling-Fans/Commercial-Electric/N-5yc1vZc4lwZdn for example

PR #107, opened 2026-04-03
fix: allow null viewport to disable viewport emulation

kitiho:
## Problem

Passing `viewport: null` to `launchPersistentContext()` has no effect — the browser always launches with `DEFAULT_VIEWPORT` (1920×947) because the `??` operator treats `null` the same as `undefined`.

## Root cause

```ts
viewport: options.viewport ?? DEFAULT_VIEWPORT
```

`??` coalesces both `null` and `undefined`, so an explicit `null` (which in Playwright means "disable viewport emulation and let the OS window control the size") is silently replaced by `DEFAULT_VIEWPORT`.

## Fix

Use a strict `undefined` check so `null` is passed through to Playwright:

```ts
viewport: options.viewport === undefined ? DEFAULT_VIEWPORT : options.viewport,
```

Also update the `viewport` type in `LaunchContextOptions` to accept `null`:

```ts
viewport?: { width: number; height: number } | null;
```

## Why it matters

Desktop apps (e.g. Electron) need the browser content to resize with the window. With `viewport: null`, Playwright/Chromium uses the actual window dimensions and responds to resize events. Without this fix there is no way to opt out of the fixed 1920×947 default.

### DIFF
--- a/js/src/playwright.ts
+++ b/js/src/playwright.ts
@@ -168,7 +168,7 @@ export async function launchPersistentContext(
       ? { proxy: typeof options.proxy === "string" ? parseProxyUrl(options.proxy) : options.proxy }
       : {}),
     ...(options.userAgent ? { userAgent: options.userAgent } : {}),
-    viewport: options.viewport ?? DEFAULT_VIEWPORT,
+    viewport: options.viewport === undefined ? DEFAULT_VIEWPORT : options.viewport,
     ...(options.colorScheme ? { colorScheme: options.colorScheme } : {}),
     ...options.launchOptions,
   });


Cloak-HQ:
Thanks for the clean PR, good catch on the `??` vs `null` behavior.

We confirmed the bug — `viewport: null` is a valid Playwright API value (disables viewport emulation), and `??` silently swallows it.

One thing: `launchPersistentContext` at line 171 has the same pattern:

```ts
viewport: options.viewport ?? DEFAULT_VIEWPORT,
```

Happy to merge this as-is and fix it ourselves, or if you'd like to add it to this PR that works too. Up to you.

kitiho:
Thanks for pointing that out! I've applied the same fix to `launchPersistentContext` as well — changed `??` to an explicit `=== undefined` check so `viewport: null` is correctly passed through there too.

Cloak-HQ:
Thanks for the fix, @kitiho! Merged and shipped.

We also applied the same fix to the Python wrapper (`browser.py`) in 1060772 — Python's Playwright API uses `no_viewport=True` instead of `viewport=None`, so we translate accordingly. You're credited as co-author on that commit.

Added you to the Contributors section in README as well.

Issue #108, opened 2026-04-03
Feature Request: Isolated World Evaluation (Undetectable DOM Access)


fmocking:
# Problem

`page.evaluate()` runs in the main world, making all automation-initiated DOM access detectable by anti-bot systems.

### Verified Detection Vectors (tested on CloakBrowser v0.3.19)

**1. Error.stack trace leaks**
```
Error: test
    at eval (eval at evaluate (:302:30), <anonymous>:3:13)
    at Object.evaluate (<anonymous>:304:16)
```
The `eval at evaluate (:302:30)` frame reveals CDP-initiated code execution. Anti-bot systems checking `Error.stack` patterns detect this.

**2. DOM mutations triggered**
`MutationObserver` on the page fires when evaluate modifies the DOM. Confirmed: `DOM mutations detected: true` after evaluate.

**3. Performance entries**
`performance.getEntries()` shows evaluate-related entries visible to page JS.

**4. addInitScript visibility**
Variables set via `page.addInitScript()` are visible to page JS in the main world.

**5. locator.toString`. native functions still report `[native code]`

## Proposed Solution

An isolated execution context for DOM reads that:
1. Can query the DOM (querySelector, getBoundingClientRect, elementFromPoint)
2. 2. Produces clean Error.stack traces (no `eval at evaluate` frames)
3. 3. Doesn't trigger MutationObservers for read-only operations
4. 4. Variables set in isolated context are NOT visible to main world JS
5. 5. No performance timeline artifacts
### API

```typescript
// Isolated. undetectable DOM read
const rect = await page.evaluateIsolated(() => {
  return document.querySelector('#btn').getBoundingClientRect();
});

// Main world. existing behavior (for when you need to modify the page)
await page.evaluate(() => { ... });
```

### Reference: Camoufox

[Camoufox](https://camoufox.com/python/main-world-eval/) (Firefox/Gecko) provides this. their `page.evaluate()` runs in an isolated world by default. Main world access requires explicit `mw:` prefix and is documented as detectable. This is the model to follow.

## Implementation

Chromium's `blink::DOMWrapperWorld` already supports multiple isolated worlds (used by Chrome extensions). A new world type for automation. separate from extension worlds to avoid extension-detection heuristics. could provide DOM access with clean execution traces.

## Environment

- CloakBrowser v0.3.19
- Chromium 145
 -  macOS arm64

Cloak-HQ:
Thanks for the detailed writeup.

[...]

PR #111, opened 2026-04-05
feat(human): implement CDP Isolated Worlds and trusted keyboard events (fixes #110)

### LINKED ISSUE #110: humanize: page.evaluate leaks in click and keyboard paths
Following up on #108 — you pointed us to CDP isolated worlds for undetectable DOM reads, and that works perfectly. However, the `humanize: true` pipeline itself uses `page.evaluate()` in three places that are detectable by the same Error.stack vector from #108.

## Verified on v0.3.19

We tested with a page that monkey-patches `document.querySelector` and checks `Error.stack` for the `:302:` evaluate frame (same detection vector anti-bot systems like Akamai use). We also monitored `isTrusted` on keyboard events.

### What's clean

**Mouse movement** (`human/mouse.js`) — fully clean. Uses `raw.move()` → CDP `Input.dispatchMouseEvent`. All events have `isTrusted=true`, no evaluate involved.

**Scrolling** (`human/scroll.js`) — fully clean. Uses `page.locator(selector).first().boundingBox()` which is a CDP call (`DOM.getBoxModel`), not evaluate.

### What leaks

**1. `isInputElement()` — `human/index.js:27`**
```javascript
async function isInputElement(page, selector) {
    return page.evaluate((sel) => {
        const el = document.querySelector(sel);
        if (!el) return false;
        const tag = el.tagName.toLowerCase();
        return tag === 'input' || tag === 'textarea'
            || el.getAttribute('contenteditable') === 'true';
    }, selector).catch(() => false);
}
```
Called on every `click()`, `dblclick()`, `type()`, and `fill()`. A page that proxies `document.querySelector` sees the selector and the `:302:` stack frame.

**Test output:**
```
click(button):  querySelector from evaluate: 1 → ["#btn"]
click(input):   querySelector from evaluate: 1 → ["#inp"]
```

**2. `isSelectorFocused()` — `human/index.js:37`**
```javascript
async function isSelectorFocused(page, selector) {
    return page.evaluate((sel) => {
        const el = document.querySelector(sel);
        return el === document.activeElement;
    }, selector).catch(() => false);
}
```
Called in `clear()`. Same querySelector + activeElement access from evaluate context.

**3. `typeShiftSymbol()` — `human/keyboard.js:87`**
```javascript
async function typeShiftSymbol(page, raw, ch, cfg) {
    await raw.down('Shift');
    await sleep(randRange(cfg.shift_down_delay));
    await raw.insertText(ch);
    await page.evaluate((key) => {
        const el = document.activeElement;
        if (el) {
            el.dispatchEvent(new KeyboardEvent('keydown', { key, bubbles: true }));
            el.dispatchEvent(new KeyboardEvent('keyup', { key, bubbles: true }));
        }
    }, ch);
    await sleep(randRange(cfg.shift_up_delay));
    await raw.up('Shift');
}
```
This has two issues:
- `page.evaluate` leaks `:302:` in Error.stack (same as #1)
- The dispatched `KeyboardEvent` has **`isTrusted=false`** — trivially detectable by any page checking `event.isTrusted`

**Test output:**
```
type("!@#$%^&*()"):
  untrusted keyboard events: 20  (all shift symbols)
  evaluate-context keyboard events: 20
```

## Reproduction

```javascript
// Set up detection on page BEFORE any humanize interaction
await page.evaluate(() => {
    const origQS = document.querySelector.bind(document);
    document.querySelector = function(sel) {
        try { throw new Error(); } catch (e) {
            if (e.stack.includes(':302:')) {
                console.log('[DETECTED] querySelector from evaluate:', sel);
            }
        }
        return origQS(sel);
    };

    document.addEventListener('keydown', (e) => {
        if (!e.isTrusted) console.log('[DETECTED] Untrusted keydown:', e.key);
    }, true);
});

// These trigger detections:
await page.click('#btn');           // → querySelector("#btn") from evaluate
await page.keyboard.type('Hello!'); // → isTrusted=false on "!"
```

## Suggested fixes

**For `isInputElement` and `isSelectorFocused`:** These are DOM reads — the exact use case for CDP isolated worlds from #108. Could use `Page.createIsolatedWorld` + `Runtime.evaluate` to check tagName and activeElement without the `:302:` stack leak.

**For `typeShiftSymbol`:** The synthetic `KeyboardEvent` via evaluate is the harder one. CDP `Input.dispatchKeyEvent` with `type: 'keyDown'` + the `key` field produces `isTrusted=true` events. The `insertText` call already handles the actual character insertion — the synthetic keydown/keyup might only be needed for sites that listen for the key events specifically.

## Environment

- CloakBrowser v0.3.19
- Chromium 145
- macOS arm64
- Bun 1.3.10

evelaa123:
### feat(human): implement CDP Isolated Worlds and trusted keyboard events (fixes #110)

This PR provides a comprehensive fix for the detection vectors identified in Issue #110 and #108. The implementation transitions the `humanize` layer from high-level synthetic interactions to native-level browser operations using the Chrome DevTools Protocol (CDP).

#### 1. The Problem: "The Evaluate Leak"
Previously, any DOM check (like `isInputElement` or `isSelectorFocused`) used standard `page.evaluate()`. 
- **The Catch:** Modern anti-bot solutions (Akamai, Cloudflare) monkey-patch `document.querySelector` or trigger errors to inspect the `Error.stack`. 
- **The Result:** The presence of specific Playwright/Chromium internal frames (e.g., `:302:`) in the stack trace instantly flags the session as automated.

#### 2. The Solution: CDP Isolated Worlds
I've introduced a **Stealth Evaluation Infrastructure** (`StealthEval` in JS, `_SyncIsolatedWorld` / `_AsyncIsolatedWorld` in Python).
- **How it works:** All internal DOM queries now execute in a dedicated **Isolated World** via `Page.createIsolatedWorld`. 
- **Why it matters:** - Scripts in the Main World cannot see or intercept these calls.
  - The `Error.stack` remains perfectly clean, as if the calls originated from a browser extension or native code.
  - Contexts are automatically invalidated and recreated on navigation to prevent stale ID errors.

#### 3. Native Keyboard Emulation (Trusted Events)
Typing shift-symbols (like `!`, `@`, `#`) previously relied on synthetic JS events.
- **The Problem:** Synthetic events carry `isTrusted: false`, which is a primary signal for modern login forms to block input.
- **The Fix:** Implemented a mapping for 21 standard shift symbols to their physical `code` and `windowsVirtualKeyCode`. These are now dispatched via `Input.dispatchKeyEvent`.
- **The Result:** Events are indistinguishable from hardware keyboard input, carrying the `isTrusted: true` flag and correct modifier bits.

#### 4. Detailed Changes:
- **JS/TS:** Refactored `index.ts` to include `StealthEval` class; patched `page` and `frame` methods to route checks through the isolated world.
- **Python:** Added sync/async isolated world managers in `__init__.py`; updated `keyboard.py` and `keyboard_async.py` with CDP dispatch logic.
- **Fallbacks:** Implemented a robust fallback mechanism—if a CDP session cannot be established, the system reverts to standard evaluation to maintain functionality.

#### 5. Verification & Testing:
Included a new reproduction suite `test_stealth_reproduction_110.py` that mirrors the detection logic used by top-tier anti-bot providers.

**Test Results:**
- `querySelector` stack trace leak: **Fixed (0 detections)**
- `isTrusted` keyboard flag: **Fixed (All events Trusted)**
- `Shift` symbol sequence: **Verified (Realistic keydown/keyup flow)**

---
*Note: This PR also includes `js/tests/stealth.test.ts` to ensure the TypeScript implementation remains stable in future releases.*

### DIFF
--- a/cloakbrowser/human/__init__.py
+++ b/cloakbrowser/human/__init__.py
@@ -3,14 +3,20 @@
 Activated via humanize=True in launch() / launch_async().
 Patches page methods to use Bezier mouse curves, realistic typing, and smooth scrolling.
 
+Stealth-aware (fixes #110):
+  - isInputElement / isSelectorFocused use CDP Isolated Worlds instead of page.evaluate
+  - Shift symbol typing uses CDP Input.dispatchKeyEvent for isTrusted=true events
+  - Falls back to page.evaluate only when CDP session is unavailable
+
 Supports both sync and async Playwright APIs.
 """
 
 from __future__ import annotations
 
+import json
 import logging
 import sys
-from typing import Any
+from typing import Any, Optional
 
 from .config import HumanConfig, HumanPreset, resolve_config
 from .config import rand, rand_range, sleep_ms, async_sleep_ms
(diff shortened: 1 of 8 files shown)


Cloak-HQ:
Thanks for the detailed PR and the thorough test coverage.

We reproduced all detection vectors from #110 on v0.3.19 with `humanize=True`:

- **3 querySelector evaluate leaks** — `_is_input_element()` and `_is_selector_focused()` expose the evaluate frame via `page.evaluate()`
- **22 untrusted keyboard events** — every shift symbol (`!@#$%^&*()`) fires `isTrusted=false` keydown + keyup from `_type_shift_symbol()`

Your keyboard fix (CDP `Input.dispatchKeyEvent` with proper key codes and modifier flags) is solid.

Before we merge, a few things to clean up:

1. **World name** — `__cloak_human` in `Page.createIsolatedWorld` is fingerprintable. Anti-bot scripts can enumerate execution contexts. Please use an empty string `""` or a randomized name.
2. **Remove the `package-lock.json` changes** — the version bump (0.3.9 → 0.3.19) and bin field are unrelated to this PR
3. **Remove the `config.ts` whitespace-only changes** — keeps the diff focused
4. **Duplicate comment block in `js/tests/stealth.test.ts`** around line 1680 — copy-paste artifact

One heads-up: as discussed in #108, we're planning to add a public `page.stealth_evaluate(expression)` API that wraps the isolated world lifecycle into a standalone module. The humanize layer will then use it internally for DOM checks, and users will also be able to call it directly for their own scripts. So expect some refactoring of the `_SyncIsolatedWorld` / `_AsyncIsolatedWorld` / `StealthEval` classes after merge into a shared module. Just don't want that to be a surprise.

Once you push the cleanup, we'll merge. Thanks again for the contribution.

evelaa123:
Thanks for the review! All four items addressed:

1. **worldName** — changed from `"__cloak_human"` to `""` in both Python (sync + async) and TypeScript
2. **package-lock.json** — reverted, no longer in the diff
3. **config.ts** — reverted whitespace-only changes
4. **Duplicate comment** — removed from `stealth.test.ts`

Cloak-HQ:
Merged, thanks! All tests pass on our end (Python 316 passed, JS 173 passed).

We'll follow up with the `stealth_evaluate` module extraction in a separate PR as mentioned above.

PR #129, opened 2026-04-09
feat: Add Puppeteer humanize support and fix Playwright humanize gaps

evelaa123:
<!DOCTYPE html>
<html>
<body>

<h2>Fix critical Playwright humanize gaps + Add full Puppeteer support</h2>

<h3>Overview</h3>
<p>Found and fixed <strong>3 critical stealth bypasses</strong> in the existing Playwright humanize implementation, then added <strong>complete Puppeteer humanize support</strong> mirroring the same architecture.</p>

<hr>

<h3>Issues Found & Fixed</h3>

<h4>Issue #1 (CRITICAL): 3 page-level methods not patched in Playwright</h4>
<p><strong>File:</strong> <code>js/src/human/index.ts</code></p>

<p><strong>Problem:</strong> Functions <code>humanPressSequentiallyFn</code>, <code>humanTapFn</code>, <code>humanClearFn</code> were implemented but never assigned to <code>page.pressSequentially</code>, <code>page.tap</code>, <code>page.clear</code>. They only existed as internal properties for frame patches.</p>

<p><strong>Impact:</strong></p>
<ul>
<li><code>page.pressSequentially('input', 'text')</code> bypassed humanize completely, typing instantly with <code>isTrusted=false</code> for shift characters</li>
<li><code>page.tap('button')</code> bypassed Bézier mouse movement, performing instant clicks</li>
<li><code>page.clear('input')</code> bypassed CDP Isolated World focus checks</li>
</ul>

<p>This directly compromised stealth. Anti-bot systems could detect superhuman typing speed and instant mouse teleportation.</p>

<p><strong>Fix:</strong> Added 3 missing page-level patches</p>

<hr>

<h4>Issue #2 (CRITICAL): 2 frame-level methods not patched in Playwright</h4>
<p><strong>File:</strong> <code>js/src/human/index.ts</code>, function <code>patchSingleFrame()</code></p>

<p><strong>Problem:</strong> Frame-level patching covered 11 methods but was missing <code>pressSequentially</code> and <code>tap</code>.</p>

<p><strong>Why this matters:</strong> Users commonly write code like this:</p>
<pre>
await page.locator('#input').pressSequentially('Hello!');
await page.locator('#button').tap();
</pre>

<p>Locator calls go through Frame methods. Without frame-level patches, the entire Locator chain bypasses humanize even if page-level patches exist.</p>

<p><strong>Fix:</strong> Added 2 missing frame-level patches</p>

<hr>

[...]

PR #133, opened 2026-04-10
feat(humanize): add Playwright ElementHandle support and fix async tests

evelaa123:
### Summary
- **ElementHandle Support:** Now `query_selector` (Python) and `$` (JS) return patched handles. Interactions like `el.click()`, `el.fill()`, and `el.hover()` use humanized Bézier curves and realistic timing.
- **Full Parity:** Works across JavaScript and Python (both Sync and Async APIs).
- **Bug Fixes:** 
    - Fixed the Python test suite (missing `asyncio` imports and `asyncio.run` conflicts).
    - Fixed a crash when passing an empty string as a proxy. (in test)
- **Status:** All 61 Python tests and 91 JS tests are passing.

### How it works
- **Automatic Patching:** The library intercepts `query_selector`, `query_selector_all`, and `wait_for_selector`. Any handle they return is automatically wrapped with humanized methods.
- **Shared State:** All handles share the same cursor position with the `Page`, so movement is continuous and natural.
- **Recursive:** If you search for a child element inside an `ElementHandle`, that child is also automatically humanized.

### Note
I've verified everything with the visual test suite. Since I haven't worked extensively with Playwright's `ElementHandle` internals before, please double-check the patching logic.

Having real good time doing this. Learning smth new

### DIFF
--- a/cloakbrowser/human/__init__.py
+++ b/cloakbrowser/human/__init__.py
@@ -900,6 +900,9 @@ def patch_page(page: Any, cfg: HumanConfig, cursor: _CursorState) -> None:
     # --- Patch Frame-level methods (for sub-frames) ---
     _patch_frames_sync(page, cfg, cursor, raw_mouse, raw_keyboard, originals)
 
+    # --- Patch ElementHandle selectors (query_selector, query_selector_all, wait_for_selector) ---
+    _patch_page_element_handles_sync(page, cfg, cursor, raw_mouse, raw_keyboard, originals, stealth, cdp_session)
+
     # Initialize cursor immediately so it doesn't visibly jump from (0,0)
     cursor.x = rand(cfg.initial_cursor_x[0], cfg.initial_cursor_x[1])
     cursor.y = rand(cfg.initial_cursor_y[0], cfg.initial_cursor_y[1])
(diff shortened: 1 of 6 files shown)


Cloak-HQ:
Thanks for this, great addition! One question on the `el.focus()` behavior.

The humanized `el.focus()` does `_move_to_element()` + `human_click()`, while stock Playwright `el.focus()` is programmatic (no click). We ran a side-by-side test:

| Scenario | Stock Playwright | Humanized (PR) |
|----------|-----------------|-----------------|
| `focus()` on `<input>` | `['input-focused']` | `['input-focused', 'input-clicked']` |
| `focus()` on `<button type=submit>` | focus only, no submit | **form submitted** |
| `focus()` on button with `onclick` | focus only, panel stays closed | **panel toggled** |
| `focus()` on `<a href>` | `['link-focused']` | JS context lost (navigation attempt) |

Was the click intentional (human-like "focus by clicking")? Or should it move the cursor then call `_orig_focus()` to keep the programmatic semantics? Something like:

```python
def _human_el_focus() -> None:
    _move_to_element()  # human-like cursor movement
    _orig_focus()       # programmatic focus, no click side-effects
```

evelaa123:
My bad, the click in focus() wasn't intentional. Fixed — now it does moveToElement() + origFocus() exactly as you suggested. Updated in Python sync/async and JS.

Cloak-HQ:
Merged, thanks! All tests pass including integration after rebase.

One note for a follow-up: `moveToElement()` in both JS and Python calls `boundingBox()` without scrolling the element into view first. If the element is below the fold, `boundingBox()` returns null and it falls back to the raw unpatched method with no humanization.

The page-level click path handles this correctly via `scrollToElement()`. The ElementHandle path should do something similar. Same gap exists in the Puppeteer ElementHandle from #129. Would be great to fix both together.

evelaa123:
No problem, will be fixid soon. Thanks for details!

PR #151, opened 2026-04-12
fix(types): type humanConfig properly

eofreternal:
Previously, `humanConfig` wasn't typed properly. It had a type of `any` so it was hard to tell what could be changed with `humanConfig`. This adds types to fix that.


Cloak-HQ:
Thanks for the PR! Merged. Clean improvement — the `as any` casts in the JS wrapper were overdue for removal.

PR #177, opened 2026-04-26
feat(lambda): add AWS Lambda one-shot scrape integration

AlexTech314:
CloakHQ team,

Hope all is well! I am obsessed with this project. Thank you for making it open source.

I'd like to offer an integration example -

### Overview

A self-contained integration that runs CloakBrowser stealth Chromium one-shot scrapes inside an AWS Lambda function (container image package type). The image is `FROM cloakhq/cloakbrowser` on Docker Hub, so every other surface from the canonical image (python, cloakserve, cloaktest, node, bash, examples) keeps working. Lambda is just an additional invocation surface.

Four files, no dependencies outside the directory:

- `Dockerfile` layers `awslambdaric`, the `aws-lambda-rie` binary for local testing, the dual-mode entrypoint, and the handler on top of the canonical image.
- `lambda-entrypoint.sh` starts Xvfb on `:99`, then routes `module.func` CMDs through `awslambdaric` (RIE locally, AWS Runtime API in production) and execs everything else directly.
- `lambda_handler.py` is the default CMD. Takes `{url, proxy?, humanize?, wait_until?, wait_for_selector?, wait_ms?, smart_wait?, ...}` and returns `{title, url, html, screenshot_b64?}`. Always headed via Xvfb.
- `INSTRUCTIONS.md` covers build, local smoke-test, event schema, and config recommendations. Deployment-tool agnostic (CLI, CDK, Terraform, SAM, console).

Two Lambda-specific Chromium flags are baked into the handler:

- `--no-zygote`: Lambda's restricted process model can't fork from Chromium's zygote, so without this child renderers fail to spawn and `new_page()` raises `TargetClosedError`.
- `--disable-dev-shm-usage`: Lambda's `/dev/shm` is ~64 MB.

Default wait is `smart_wait`: polls `document.outerHTML.length` and bails when stable for `dom_stable_ms` (default 1500, capped at `max_settle_ms` 15000). More robust than `wait_until=networkidle` for at-scale scraping because it ignores network activity that doesn't mutate the DOM (analytics beacons, long-poll, websockets). Standard wait primitives still work and override `smart_wait`.

### Testing

- All canonical image surfaces preserved: REPL, examples, cloakserve, cloaktest, node, bash.
- End-to-end Lambda scrape verified against a Next.js SPA with lazy-loaded emails (8/8 captured warm), `example.com` (~3 s warm), and `news.ycombinator.com/news` (30/30 stories, ~4 s warm).
[...]

PR #180, opened 2026-04-28
feat(lambda): cold-start hardening + handler-side retry orchestration

AlexTech314:
CloakHQ team,

Quick follow-up to #177. While running the Lambda integration at scale (a private 3454-site sample) I hit two related production issues that the merged version didn't catch. Fixes are scoped to the integration directory only.

### Overview

Two changes, both inside `examples/integrations/aws_lambda/`:

**1. Cold-start hardening** (`lambda-entrypoint.sh` + `lambda_handler.py`)

Concurrent cold-start storms expose a race in the Xvfb startup. The merged version's `sleep 0.5` is too short under CPU contention, and a previous Xvfb sometimes dies and leaves a stale `/tmp/.X99-lock` so the next start refuses with "Server is already active for display 99". Both surface as `Looks like you launched a headed browser without having a XServer running` in the handler. We saw this hit ~10% of cold containers at 100-concurrent invocations and ~11% at 900-concurrent.

The entrypoint now:

- Removes `/tmp/.X99-lock` and `/tmp/.X11-unix/X99` before starting Xvfb (clean slate).
- Polls for the X11 socket to appear (up to 10s) instead of a fixed sleep.
- Adds a 200ms buffer after the socket appears so `listen()` and `accept()` finish before the runtime client takes its first invocation.

The handler adds an in-process `_launch_with_retry` (3 attempts, 0.3s + 0.6s backoff) as belt-and-suspenders. Catches anything the entrypoint fix misses; a retry on a now-warm container almost always succeeds.

**2. Handler-side retry orchestration** (`lambda_handler.py`)

The merged version surfaces every transient post-launch failure (cert errors, navigation timeouts, connection timeouts) directly to the caller. In practice these are routinely recoverable inside one Lambda invocation by relaunching with adjusted Chromium args or a longer goto budget. Doing it in the handler is cheaper than asking each caller to re-implement retry orchestration on top, and it keeps a recovery on the same billed invocation rather than a second one.

```
Strategies (priority order, first match wins):
  ERR_CERT_*                  -> --ignore-certificate-errors + 60s goto
  Timeout exceeded            -> 90s goto + 25s smart_wait cap
  ERR_CONNECTION_TIMED_OUT    -> same as Timeout

Not retried (unrecoverable):
  ERR_NAME_NOT_RESOLVED, ERR_SSL_PROTOCOL_ERROR, ERR_CONNECTION_REFUSED,
  ERR_HTTP_RESPONSE_CODE_FAILURE  (these bail immediately)
```

A new event field `retries` (default 1, set to 0 to disable) bounds the number of strategy retries. The success response shape is unchanged. On final failure, the raised `RuntimeError` includes a `retry_history` block listing every attempt that was tried.

`INSTRUCTIONS.md` documents the new event field, lists the strategies, and bumps the function timeout recommendation from 60-120s to 120-180s. The 90s timeout-class retry budget plus cleanup means a single invocation can now span up to ~130s; 180s leaves headroom.

### Testing

Same 3454-site sample (seed=1), measured across iterations of this work:

```
v1 baseline (no fixes, c=100):                          13.5% failure rate, $1.07
v2 (entrypoint Xvfb poll only, c=100):                   9.9% failure rate, $1.11
v3 (cold-start fix + bench-side retry, c=900):           3.3% failure rate, $1.32
This change (handler-side retry, no bench retry, c=250): 2.1% failure rate, $1.13
```

The final 2.1% are all genuinely unrecoverable. Mostly `ERR_NAME_NOT_RESOLVED` (DNS doesn't exist), `ERR_SSL_PROTOCOL_ERROR` (broken HTTPS at the site), `ERR_CONNECTION_REFUSED` (site down), `ERR_HTTP_RESPONSE_CODE_FAILURE` (4xx/5xx from the server), and a handful of "response payload exceeded 6 MB" hits against the Lambda invoke-response limit.

Notable: zero `ERR_CERT_*` failures and zero "launch failed" failures across the 3454-site batch. Both classes that previously dominated the failure tail are gone.

This run is also cheaper than v3 despite the better outcomes ($1.13 vs $1.32) because cert/timeout retries now happen inline within one Lambda invocation instead of as a second invocation.

The `retries: 0` opt-out and the unrecoverable-error bail-out were both verified locally with the bundled RIE.

Happy to iterate. The change is intentionally scoped to the integration directory and is purely additive for callers who don't pass `retries`.

Best,
Alex S.

### DIFF
--- a/examples/integrations/aws_lambda/INSTRUCTIONS.md
+++ b/examples/integrations/aws_lambda/INSTRUCTIONS.md
@@ -113,6 +113,29 @@ Only `url` is required. Everything else is optional.
 | `screenshot` | bool | `true` |
 | `full_page_screenshot` | bool | `false` |
 
+### Retry orchestration
+
+The handler retries transient navigation failures inline within the same Lambda invocation. Two layers, both built-in:
+
+- **Launch retries** — 3 attempts with 0.3 s + 0.6 s backoff. Recovers Xvfb / Chromium spawn races at cold start. Fast and cheap; not configurable.
+- **Strategy retries** — default 1 attempt, configurable via the `retries` event field. Recovers specific post-launch error classes by relaunching with adjusted Chromium args / page-load budgets.
+
+| Field | Type | Default |
+|---|---|---|
+| `retries` | int | `1` — number of strategy-retry attempts after the first failure. Set to `0` to disable retry entirely. |
+
+Strategies (priority order — first match wins):
+
+| Error pattern | Strategy applied |
+|---|---|
+| `ERR_CERT_*` (any cert error) | `extra_args: ["--ignore-certificate-errors"]`, `goto_timeout_ms: 60000` |
+| `Timeout … exceeded` | `goto_timeout_ms: 90000`, `max_settle_ms: 25000` |
+| `ERR_CONNECTION_TIMED_OUT` | same as `Timeout … exceeded` |
+
+Errors that are **not retried** (no anonymous scraper can recover): `ERR_NAME_NOT_RESOLVED`, `ERR_SSL_PROTOCOL_ERROR`, `ERR_CONNECTION_REFUSED`, `ERR_HTTP_RESPONSE_CODE_FAILURE`. These bail immediately.
+
+On final failure, the raised `RuntimeError`'s message includes a `retry_history` block listing every attempt (strategy applied + error seen). Successful invocations return the standard response shape unchanged — no surprise fields when retries didn't fire.
+
 ### Response
 
 ```json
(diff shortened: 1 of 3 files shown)


Cloak-HQ:
Thanks for the follow-up, Alex. The cold-start data is really convincing (13.5% → 2.1%), and the handler-side retry design is the right call — keeping retries on the same billed invocation is cleaner for callers and cheaper overall.

Entrypoint hardening and launch retry both look good. One bug to fix in the strategy retry path before we merge:

**`extra_args` gets clobbered on cert retry**

In `_run`, the strategy merge is a flat dict spread:

```python
current_event = {**current_event, **strategy}
```

The cert strategy returns `{"extra_args": ["--ignore-certificate-errors"], ...}`, which **replaces** the caller's `extra_args` instead of appending. If someone passes `extra_args: ["--proxy-server=http://..."]` or any custom flags, those are silently lost on the cert retry.

Fix (in `_run`, replacing the merge line):

```python
merged_args = list(current_event.get("extra_args", [])) + strategy.get("extra_args", [])
current_event = {**current_event, **strategy, "extra_args": merged_args}
```

This keeps strategies as pure data and puts the merge logic in the orchestrator where it belongs.

Everything else is clean. The expanded diag coverage (now fires on all final failures, not just launch) is a nice side-effect of the refactor.

AlexTech314:
Great catch!

Cloak-HQ:
Thanks for iterating on this, Alex. Merged!

The `extra_args` fix looks good. Let us know how it holds up on your next batch run.

PR #201, opened 2026-05-08
fix(js): preserve iframe scope in humanized frame actions

### LINKED ISSUE #184: Humanization: Frame action patches delegate to Page actions, causing loss of iframe selector scope.
Description: 

Cloak's JavaScript humanization layer patches `Frame` action methods by delegating them to the owning `Page` action methods. This appears to break iframe-scoped interactions.

Expected behavior:

`frame.click(selector)`, `frame.fill(selector, value)`, and Locator actions created from a frame should preserve the original Playwright frame scope. A selector inside an iframe should be resolved inside that iframe.

Actual behavior:

In `patchSingleFrame`, frame methods delegate to page methods:

```ts
(frame as any).click = async (selector, options) => {
  await (page as any).click(selector, options);
};

(frame as any).fill = async (selector, value, options) => {
  await (page as any).fill(selector, value, options);
};

(frame as any).type = async (selector, text, options) => {
  await (page as any).type(selector, text, options);
};
```

In Playwright, `page.click(selector)` delegates to the main frame only, while `frame.click(selector)` is scoped to that frame. So after Cloak patches a child frame, iframe-contained selectors may be searched in the main frame instead of the iframe.

This also affects frame-owned Locators, because Playwright Locator methods call their owning frame:

```js
locator.click() -> this._frame.click(this._selector, ...)
locator.fill()  -> this._frame.fill(this._selector, value, ...)
```

After Cloak patches `frame.click`, that frame-owned Locator effectively becomes:

```js
iframeLocator.click() -> frame.click(selector) -> page.click(selector)
```

That loses iframe scope.

CloakBrowser version: 

```txt
cloakbrowser@0.3.26
```

Also checked PR/commit:

```txt
PR #183
feat: per-call human_config overrides, timeout forwarding, humanized scrollIntoViewIfNeeded
merge commit: 661b873
```

The frame delegation pattern still appears present there.

Wrapper: 

```txt
JavaScript / Playwright
```

Environment: 

```txt
macOS
Not using Docker
Node.js / Playwright JavaScript wrapper
```

Launch options:

```js
import { launchPersistentContext } from 'cloakbrowser';

const context = await launchPersistentContext({
  userDataDir: './profile-cloak'
});
```

Also relevant when applying Cloak humanization to an attached Playwright context via the exported humanization API:

```js
import { patchContext, resolveConfig } from 'cloakbrowser/human';

patchContext(context, resolveConfig('default'));
```

Tested with a different IP or proxy? 

```txt
No. This appears unrelated to network/proxy behavior.
```

Works outside Docker / on host machine? 

```txt
Not using Docker.
```

Steps to reproduce:

```html
<!doctype html>
<html>
  <body>
    <iframe
      id="form-frame"
      srcdoc='<!doctype html>
        <html>
          <body>
            <input id="inside-frame" value="">
            <button id="submit">Submit</button>
          </body>
        </html>'
    ></iframe>
  </body>
</html>
```

```js
import { launchPersistentContext } from 'cloakbrowser';

const context = await launchPersistentContext({
  userDataDir: './profile-cloak'
});

const page = await context.newPage();
await page.goto('http://127.0.0.1:3000/repro.html');

const frame = page.frame({ name: 'form-frame' }) ||
  page.frames().find(f => f.url().includes('about:srcdoc'));

await frame.fill('#inside-frame', 'hello');
await frame.click('#submit');
```

Expected:

```txt
The input and button are found inside the iframe.
The actions are humanized by Cloak while preserving frame scope.
```

Actual:

```txt
The patched frame methods delegate to page.fill/page.click.
Those page methods resolve selectors from the main frame, not the child iframe.
The iframe selector can fail or resolve the wrong main-frame element if one exists.
```

A minimal source-level reproduction of the routing issue:

```js
const { Locator } = require('playwright-core/lib/client/locator.js');

const calls = [];

const page = {
  click(selector, options) {
    calls.push({ owner: 'page.click', selector, options });
    return Promise.resolve();
  },
  fill(selector, value, options) {
    calls.push({ owner: 'page.fill', selector, value, options });
    return Promise.resolve();
  }
};

const frame = {
  _platform: {},
  _timeout() { return 0; },
  _wrapApiCall(fn) { return fn(); },
  page() { return page; },
  click(selector, options) {
    calls.push({ owner: 'frame.click', selector, options });
    return Promise.resolve();
  },
  fill(selector, value, options) {
    calls.push({ owner: 'frame.fill', selector, value, options });
    return Promise.resolve();
  }
};

(async () => {
  const locator = new Locator(frame, '#inside-frame');

  await locator.click();
  await locator.fill('before');

  // Cloak-style frame patch:
  frame.click = async (selector, options) => page.click(selector, options);
  frame.fill = async (selector, value, options) => page.fill(selector, value, options);

  await locator.click();
  await locator.fill('after');

  console.log(calls);
})();
```

Observed routing:

```json
[
  {
    "owner": "frame.click",
    "selector": "#inside-frame"
  },
  {
    "owner": "frame.fill",
    "selector": "#inside-frame",
    "value": "before"
  },
  {
    "owner": "page.click",
    "selector": "#inside-frame"
  },
  {
    "owner": "page.fill",
    "selector": "#inside-frame",
    "value": "after"
  }
]
```

Error output / screenshots:

```txt
No stack trace required. The issue is visible from the patched call routing:
frame-owned actions are redirected to page-owned actions, losing iframe scope.
```

Dockerfile (if applicable):

```txt
N/A
```

Additional notes:

Why this matters:

Many real-world forms and challenge flows are iframe-heavy. For those pages, preserving frame scope is not optional. Playwright intentionally separates:

```js
page.click(selector);   // main frame
frame.click(selector);  // specific frame
```

Cloak should preserve that contract while adding humanization.

This also matters for stealth and maintainability. If users discover that iframe `Locator` actions break under Cloak, they may work around it by adding coordinate bridges or custom monkey patches. Those workarounds can introduce new fingerprints, such as deterministic click positions or inconsistent action routing.

A better fix would be for Cloak's frame patches to preserve frame selector scope while still using Cloak-owned mouse/keyboard behavior. For example, frame actions could resolve/wait/actionability-check inside the frame, then execute the final mouse/keyboard action through the Cloak humanized page driver using the element's viewport coordinates.

manaskarra:
## Summary
- Fix humanized frame actions so they preserve iframe scope by resolving selectors through `frame.locator(selector)` instead of delegating to `page.*` methods.
- Keep frame actions on the shared humanized mouse/keyboard path.
- Add regression coverage proving `frame.click()` uses `frame.locator()` and does not call the page click implementation.

Fixes #184.

## Test Plan
- `npm run typecheck`
- `npm test -- --run tests/humanize.test.ts`
- `npm test`

### DIFF
--- a/js/src/human/index.ts
+++ b/js/src/human/index.ts
@@ -607,6 +607,8 @@ function patchSingleFrame(
     await humanClick(raw, moved.isInput, moved.callCfg);
   };
 
+  const getFrameCdp = async () => stealth.getCdpSession().catch(() => null);
+
   const frameHover = async (selector: string, options?: any) => {
     const moved = await moveToFrameSelector(selector, options, false);
     if (!moved) return origFrameHover(selector, options);
@@ -629,7 +631,8 @@ function patchSingleFrame(
     await sleep(randRange(callCfg.field_switch_delay));
     await frameClick(selector, options);
     await sleep(rand(100, 250));
-    await humanType(page, rawKb, text, callCfg).catch(() => origFrameType(selector, text, options));
+    const cdp = await getFrameCdp();
+    await humanType(page, rawKb, text, callCfg, cdp).catch(() => origFrameType(selector, text, options));
   };
 
   (frame as any).fill = async (selector: string, value: string, options?: any) => {
@@ -641,7 +644,8 @@ function patchSingleFrame(
     await sleep(rand(30, 80));
     await originals.keyboardPress('Backspace');
     await sleep(rand(50, 150));
-    await humanType(page, rawKb, value, callCfg).catch(() => origFrameFill(selector, value, options));
+    const cdp = await getFrameCdp();
+    await humanType(page, rawKb, value, callCfg, cdp).catch(() => origFrameFill(selector, value, options));
   };
 
   (frame as any).check = async (selector: string, options?: any) => {
@@ -674,7 +678,8 @@ function patchSingleFrame(
       await frameClick(selector, options);
     }
     await sleep(rand(100, 250));
-    await humanType(page, rawKb, text, callCfg).catch(() => origFramePressSequentially?.(selector, text, options));
+    const cdp = await getFrameCdp();
+    await humanType(page, rawKb, text, callCfg, cdp).catch(() => origFramePressSequentially?.(selector, text, options));
   };
 
   (frame as any).tap = async (selector: string, options?: any) => {
--- a/js/src/human/index.ts
+++ b/js/src/human/index.ts
@@ -529,16 +529,37 @@ function patchFrames(
   stealth: StealthEval,
 ): void {
   for (const frame of iterFrames(page)) {
-    patchSingleFrame(frame, page, cfg, originals, stealth);
+    patchSingleFrame(frame, page, cfg, cursor, raw, rawKb, originals, stealth);
     // Patch frame-level ElementHandle selectors ($, $$, waitForSelector)
     patchFrameElementHandles(frame, page, cfg, cursor, raw, rawKb, originals, stealth);
   }
+}
+
+function firstFrameLocator(frame: Frame, selector: string): any {
+  const locator = frame.locator(selector) as any;
+  return typeof locator.first === 'function' ? locator.first() : locator;
+}
+
+async function isFrameInputElement(frame: Frame, selector: string): Promise<boolean> {
+  return firstFrameLocator(frame, selector).evaluate((el: Element) => {
+    const tag = el.tagName.toLowerCase();
+    return tag === 'input' || tag === 'textarea'
+      || el.getAttribute('contenteditable') === 'true';
+  }).catch(() => false);
+}
+
+async function isFrameSelectorFocused(frame: Frame, selector: string): Promise<boolean> {
+  return firstFrameLocator(frame, selector).evaluate((el: Element) => el === document.activeElement)
+    .catch(() => false);
 }
 
 function patchSingleFrame(
   frame: Frame,
   page: Page,
   cfg: HumanConfig,
+  cursor: CursorState,
+  raw: RawMouse,
+  rawKb: RawKeyboard,
   originals: any,
   stealth: StealthEval,
 ): void {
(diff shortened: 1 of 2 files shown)


Cloak-HQ:
Thanks for the PR — the iframe scope fix is correct and well-tested. Frame actions should resolve selectors through `frame.locator()`, not delegate to `page.*`.

One thing to fix before we merge:

**Missing CDP session in `humanType` calls**

Page-level type/fill/pressSequentially pass the CDP session for stealth shift-key typing (`Input.dispatchKeyEvent` with `isTrusted=true`). The frame versions omit it:

```ts
// page (stealth path — lines 366, 380, 440)
await humanType(page, rawKb, text, callCfg, cdp);

// frame (lines 632, 644, 677 — no cdp, falls back to detectable page.evaluate)
await humanType(page, rawKb, text, callCfg).catch(...)
```

The CDP session is page-level (not frame-scoped), so it can be passed through directly. Without it, shift symbols (`@`, `#`, `!`, `$`, etc.) typed into iframe fields fall back to `page.evaluate` which produces detectable stack traces.

Affects `frame.type()`, `frame.fill()`, and `frame.pressSequentially()`.

manaskarra:
> Thanks for the PR — the iframe scope fix is correct and well-tested. Frame actions should resolve selectors through `frame.locator()`, not delegate to `page.*`.
> 
> One thing to fix before we merge:
> 
> **Missing CDP session in `humanType` calls**
> 
> Page-level type/fill/pressSequentially pass the CDP session for stealth shift-key typing (`Input.dispatchKeyEvent` with `isTrusted=true`). The frame versions omit it:
> 
> ```ts
> // page (stealth path — lines 366, 380, 440)
> await humanType(page, rawKb, text, callCfg, cdp);
> 
> // frame (lines 632, 644, 677 — no cdp, falls back to detectable page.evaluate)
> await humanType(page, rawKb, text, callCfg).catch(...)
> ```
> 
> The CDP session is page-level (not frame-scoped), so it can be passed through directly. Without it, shift symbols (`@`, `#`, `!`, `$`, etc.) typed into iframe fields fall back to `page.evaluate` which produces detectable stack traces.
> 
> Affects `frame.type()`, `frame.fill()`, and `frame.pressSequentially()`.

fixed, frame type, fill, and pressSequentially now pass the page-level CDP session through to humanType, so shift symbols in iframe fields stay on the Input.dispatchKeyEvent path instead of falling back to page.evaluate

Cloak-HQ:
Thanks for the fix, merged! The isolated-world improvement for frame DOM reads (so `isFrameInputElement`/`isFrameSelectorFocused` use CDP instead of `locator.evaluate`) would be a nice follow-up if you're interested.

PR #205, opened 2026-05-08
feat: add types to `options` argument

eofreternal:
Typed the remaining human interaction `options` in `js/src/human/index.ts` using `Partial<HumanConfig> & ({ timeout: number })` style, and updated the wrappers to merge config directly from `options` instead of nested `human_config`.

### DIFF
--- a/js/src/human-puppeteer/index.ts
+++ b/js/src/human-puppeteer/index.ts
@@ -336,10 +336,11 @@ function patchPage(page: Page, cfg: HumanConfig, cursor: CursorState): void {
     clickCount?: number;
     count?: number;
     delay?: number;
+    human_config?: Partial<HumanConfig>;
     timeout?: number;
   })) => {
     await ensureCursorInit();
-    const callCfg = mergeConfig(cfg, options);
+    const callCfg = mergeConfig(cfg, options?.human_config ?? options);
     if (callCfg.idle_between_actions) {
       await humanIdle(raw, cursor.x, cursor.y, callCfg);
     }
@@ -365,9 +366,9 @@ function patchPage(page: Page, cfg: HumanConfig, cursor: CursorState): void {
   };
 
   // ==== hover ====
-  const humanHoverFn = async (selector: string, options?: Partial<HumanConfig> & ({ timeout?: number })) => {
+  const humanHoverFn = async (selector: string, options?: Partial<HumanConfig> & ({ timeout?: number; human_config?: Partial<HumanConfig> })) => {
     await ensureCursorInit();
-    const callCfg = mergeConfig(cfg, options);
+    const callCfg = mergeConfig(cfg, options?.human_config ?? options);
     if (callCfg.idle_between_actions) {
       await humanIdle(raw, cursor.x, cursor.y, callCfg);
     }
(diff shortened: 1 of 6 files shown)


Cloak-HQ (commented):
Thanks for the PR! Replacing `any` with proper types is a solid improvement.

A few things to address before merging:

**Bug: `timeout` is required instead of optional (Playwright side)**

In `js/src/human/index.ts`, the type uses `{ timeout: number }` instead of `{ timeout?: number }`. This affects `click`, `dblclick`, `hover`, `type`, `fill`, `clear`, `check`, `uncheck`, `selectOption`, `press`, `pressSequentially`, `tap`, and all frame equivalents. If a user passes `{ typing_delay: 30 }` without `timeout`, TypeScript will error.

The Puppeteer file has it correct with `timeout?: number` — just needs the same fix on the Playwright side.

**Silent breakage: `human_config` per-call overrides stop working**

The PR changes from `mergeConfig(cfg, options?.human_config)` to `mergeConfig(cfg, options)`. If someone is using the current API:

```typescript
await page.type("#email", "hi", { human_config: { typing_delay: 30 } });
```

The overrides silently stop working — `human_config` gets spread as a key on the config object instead of being read as nested overrides. No error, just ignored.

Could you keep backward compat? Something like:

```typescript
const callCfg = mergeConfig(cfg, options?.human_config ?? options);
```

That way both the new flat style and the existing `human_config` style work.

**Breaking change: `humanIdle` public API**

The `seconds` parameter was removed from `humanIdle()` in `mouse.ts`. This function is exported publicly from both Playwright and Puppeteer entry points. Users calling it manually lose the ability to specify a custom duration.

---

Let us know if you have any questions on any of these!

eofreternal:
changes made

Cloak-HQ (commented):
One more thing — `js/src/human/elementhandle.ts` wasn't updated with the type changes. It still uses `options?: any` and `mergeConfig(cfg, options?.human_config)` (no flat config fallback). The Puppeteer equivalent in `human-puppeteer/index.ts` was fully updated. Could you apply the same treatment to `elementhandle.ts` for consistency?

eofreternal:
Oops. I missed that. Updated it

Cloak-HQ (commented):
The latest commit breaks `tsc` — three type errors in `elementhandle.ts`:

```
src/human/elementhandle.ts(133,57): error TS2769: No overload matches this call.
src/human/elementhandle.ts(440,54): error TS2769: No overload matches this call.
src/human/elementhandle.ts(492,59): error TS2769: No overload matches this call.
```

The `waitForSelector` typed `options` can be `undefined`, but Playwright's overloads don't accept that. Quick fix — pass a fallback:

```typescript
const child = await origElWaitForSelector(selector, options ?? {});
```

Same for the other two at lines 440 and 492. Tests pass (vitest doesn't type-check), but `npm run build` fails.

eofreternal:
Fixed. So sorry for breaking everything

Cloak-HQ:
Merged manually via squash commit 80d9f7c — rebased onto main to resolve conflicts with #201 (iframe scope fix). Thanks for the contribution!

Issue #157, opened 2026-04-14
SOCKS5 with username/password auth silently falls back to direct connection


dylanpulver:
## Summary

The 0.3.24 changelog added "Native SOCKS5 proxy support — pass `proxy='socks5://user:pass@host:port'` directly. Credentials handled natively by Chrome." In practice, traffic appears to silently bypass the SOCKS5 proxy and exit via the host's direct network interface when credentials are included.

We'd love clarification on expected behavior here, or confirmation that this is a bug — happy to help test a fix.

## Environment

- `cloakbrowser@0.3.24`
- `playwright-core@1.57.0`
- Node 20 (inside a Cloud Run container)
- Chromium 146.0.7680.177.1 (CloakBrowser's bundled build)
- Linux x64
- Proxy provider: Smartproxy/Decodo dedicated ISP (`isp.decodo.com:10001`), supports SOCKS5 per their docs (confirmed via `curl -x "socks5://user:pass@isp.decodo.com:10001"` — works, exit IP matches the assigned Decodo IP)

## What we did

```ts
import { launch } from 'cloakbrowser'

const browser = await launch({
  headless: true,
  proxy: 'socks5://USER:PASS@isp.decodo.com:10001',
  args: ['--fingerprint=12345', '--window-size=1920,1080'],
  geoip: true,
  humanize: true,
  humanPreset: 'careful',
})

// Close default context to get recordVideo options applied
const defaultContext = browser.contexts()[0]
if (defaultContext) await defaultContext.close()

const context = await browser.newContext({ viewport: { width: 1920, height: 919 } })
const page = await context.newPage()
await page.goto('https://browserleaks.com/ip')
```

## What we expected

`browserleaks.com/ip` should show the Decodo exit IP (`48.44.16.178` for port 10001).

## What actually happened

- No errors at launch or during navigation
- `browserleaks.com/ip` shows the **Cloud Run egress IP** (`34.96.60.132` — Google LLC), indicating traffic went direct and bypassed the SOCKS5 proxy
- Interestingly, WebRTC leak test still reports the Decodo IP (`48.44.16.178`), so *something* in the browser is aware of the proxy — just not the main HTTPS request path
- DNS resolution also leaks to GCP internal resolvers, consistent with DNS not going through the SOCKS5 tunnel

Swapping the same URL to `http://USER:PASS@isp.decodo.com:10001` (HTTP proxy) works correctly — exit IP is Decodo. So the proxy credentials and endpoint are valid.

[...]

PR #209, opened 2026-05-09
fix(proxy): log when SOCKS5 credential auto-encoding rewrites URL (#157)

Youhai020616:
## Summary

Auto URL-encoding of SOCKS5 credentials (added in v0.3.26 to fix Chromium's `=` truncation bug from #157) currently happens silently. Users debugging proxy connectivity have no way to know the wrapper rewrote their proxy URL — the original #157 thread took 8 round-trips between reporter and maintainer to surface this exact ambiguity.

This PR adds visibility: when re-encoding actually changes the URL, emit a log so users debugging future SOCKS5 issues can see what the wrapper did.

## Changes

- **Python** (`cloakbrowser/browser.py`): `_normalize_socks_string_url` emits `logger.info` on the `cloakbrowser` logger when re-encoding changes the URL.
- **JavaScript** (`js/src/proxy.ts`): `normalizeSocksStringUrl` emits `console.debug` when re-encoding changes the URL.
- Stays silent on already-encoded inputs and credential-less URLs (no false-positive noise).
- Credentials are not included in the log message.

## Tests

3 new tests per language in the existing suites:

**Python (`tests/test_proxy.py`)**:
- `test_socks5_string_logs_info_when_reencoding`
- `test_socks5_string_silent_when_already_encoded`
- `test_socks5_string_silent_when_no_credentials`

**JavaScript (`js/tests/proxy.test.ts`)**:
- `logs debug message when SOCKS5 credentials get re-encoded`
- `stays silent when SOCKS5 URL is already encoded (no log spam)`
- `stays silent when SOCKS5 URL has no credentials`

All 58 Python tests pass (was 55, +3). All 312 JS tests pass. TypeScript typecheck clean.

## Out of scope (deliberate)

dylanpulver's docs request from #157 (mentioning URL-encoding requirement in the README) is not included here — happy to follow up with a separate docs PR if that's useful. Kept this scoped to the runtime visibility fix.

## Test plan

- [x] `pytest tests/test_proxy.py -v` — all green
- [x] `cd js && npm test` — all green
- [x] `cd js && npm run typecheck` — clean
- [x] Verified credentials don't leak into log output
- [x] Verified idempotent path (already-encoded URL) stays silent

### DIFF
--- a/cloakbrowser/browser.py
+++ b/cloakbrowser/browser.py
@@ -832,19 +832,25 @@ def _normalize_socks_string_url(url: str) -> str:
     # urlparse returns None for absent components, "" for present-but-empty.
     if parsed.username is None and parsed.password is None:
         return url
-    enc_user = quote(unquote(parsed.username), safe="") if parsed.username else ""
+    raw_user = parsed.username or ""
+    enc_user = quote(unquote(raw_user), safe="") if raw_user else ""
     # Preserve the colon separator when password component is present, even if
     # empty, so `user:@host` stays `user:@host`.
     if parsed.password is not None:
-        enc_pass = quote(unquote(parsed.password), safe="") if parsed.password else ""
+        raw_pass = parsed.password
+        enc_pass = quote(unquote(raw_pass), safe="") if raw_pass else ""
     else:
+        raw_pass = None
         enc_pass = None
     normalized = _assemble_socks_url(
         parsed.scheme, parsed.hostname or "", parsed.port,
         enc_user, enc_pass,
         parsed.path, parsed.params, parsed.query, parsed.fragment,
     )
-    if normalized != url:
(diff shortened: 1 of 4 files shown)


Youhai020616:
Pushed dbd78ee addressing the Copilot review:

The log condition now compares raw vs encoded `username`/`password` substrings instead of the full reconstructed URL, so cosmetic urlparse normalization (lowercased scheme/hostname) no longer triggers the message. JS gets the same credential-level compare for parity, even though its manual parser preserves case today.

Added one regression test per language: `socks5://USER:pass@HOST.com:1080` must stay silent.

Cloak-HQ:
Thanks for the contribution! Clean PR, well-scoped.

Credential-level comparison instead of full-URL diff was the right call. Good catch on the Copilot feedback too, `urlparse` hostname lowercasing would have caused false positives.

Merged.

PR #213, opened 2026-05-10
fix: bound GeoIP resolution so launch cannot hang

### LINKED ISSUE #207: geoip=True
With geoip=True under heavy load, the process freezes/stalls during browser startup. Turning off geoip makes the problem go away. Adding a timeout guard (e.g., 10 seconds) around the synchronous geo-IP network calls in maybe_resolve_geoip would prevent the hang

manaskarra:
## Summary
- add a bounded timeout around Python and JS GeoIP resolution
- default to 5s and allow env overrides via `CLOAKBROWSER_GEOIP_TIMEOUT_SECONDS` / `CLOAKBROWSER_GEOIP_TIMEOUT_MS`
- keep launch APIs unchanged and continue without GeoIP when resolution stalls
- add Python and JS regression tests proving timeout returns quickly

## Why
Fixes #207. When `geoip=True`, proxy/IP lookup or database fetch can stall browser startup. This keeps the existing best-effort behavior but prevents launch from hanging indefinitely.

## Test Plan
- `python3 -m pytest tests/test_geoip.py tests/test_proxy.py -q`
- `cd js && npm test -- tests/geoip.test.ts`
- `cd js && npm run typecheck`

### DIFF
--- a/cloakbrowser/browser.py
+++ b/cloakbrowser/browser.py
@@ -16,6 +16,8 @@ from __future__ import annotations
 
 import logging
 import os
+import queue
+import threading
 from typing import Any, Literal, TypedDict
 from urllib.parse import quote, unquote, urlparse, urlunparse
 
@@ -24,6 +26,8 @@ from .download import ensure_binary
 from .human.config import HumanConfigOverrides, HumanPreset
 
 logger = logging.getLogger("cloakbrowser")
+
+DEFAULT_GEOIP_TIMEOUT_SECONDS = 5.0
 
 # Sentinel to distinguish "viewport not provided" from "viewport=None" (disable emulation)
 _VIEWPORT_UNSET = object()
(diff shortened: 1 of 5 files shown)


Cloak-HQ (commented):
Thanks for tackling this — issue #207 is a real problem and the approach is solid. Found a few things to address before merging:

**1. `AbortSignal.timeout(5000)` on DB download breaks first use**

`downloadGeoipDb` fetches a ~70MB file. With the 5s GeoIP timeout applied here, first-use downloads will almost always abort. The resolution timeout should only bound the IP lookup + DB read path, not the one-time database download.

**2. Env var unit mismatch**

Python reads `CLOAKBROWSER_GEOIP_TIMEOUT_SECONDS` (seconds), JS reads `CLOAKBROWSER_GEOIP_TIMEOUT_MS` (milliseconds). Both fall back to `CLOAKBROWSER_GEOIP_TIMEOUT` but interpret it in different units. A user setting `CLOAKBROWSER_GEOIP_TIMEOUT=5` gets 5 seconds in Python and 5 milliseconds in JS. Should be one env var with consistent units.

**3. Timeout logic should live in `geoip.py`**

Our existing timeout patterns (`timeout=10.0` on IP resolution, `timeout=300.0` on DB download) and `import threading` are all in `geoip.py`. The new timeout wrapper fits there, not in `browser.py`.

**4. Daemon thread leak on timeout**

When `_call_geoip_with_timeout` times out, the spawned thread keeps running (holding a proxy connection, doing network I/O). Under heavy load with stalling proxies — the exact scenario from #207 — this accumulates orphaned threads rather than bounding resource usage.

Cloak-HQ:
Thanks for the quick revision — all four points addressed cleanly. Merged!

We added a small follow-up commit with a few fixes found during review:
- `_resolve_webrtc_args` was still calling private `_resolve_exit_ip` without timeout — switched to `resolve_proxy_exit_ip`
- Added missing `req.on("timeout")` handler on the tunneled HTTPS request in JS (socket leak on hang)
- Added `math.isfinite` check in Python timeout parsing (JS already had `Number.isFinite`)
- Recompute deadline after CONNECT succeeds in JS proxy tunnel

PR #218, opened 2026-05-11
Update security policy with reporting guidelines and scope

0xlally:


Cloak-HQ:
Thanks for the contribution! We appreciate the thought, but this security policy is a generic template that doesn't reflect CloakBrowser's actual architecture or attack surface. We're a browser wrapper, not a web service, so most of the in-scope items (auth bypass, path traversal, RCE) don't apply here.

If we add a SECURITY.md in the future, it would need to be tailored to our specific concerns. Closing for now, but thanks for the interest.

PR #222, opened 2026-05-12
fix(security): bump tar + transitive deps via npm audit fix

aaronjmars:
## Summary

Patches 8 of 14 dependency CVEs flagged by `osv-scanner` against `js/package-lock.json`. All bumps fit within the existing semver ranges in `package.json`, produced by `npm audit fix --package-lock-only`. The other 6 are gated on a `puppeteer-core` / `vitest` major-version bump and intentionally left for a separate PR.

## Impact

The only runtime (shipped-to-users) dep in the bundle is **`tar`** (7.5.9 → 7.5.15), which addresses two HIGH CVEs:

- `GHSA-9ppj-qmqm-q256` — Symlink Path Traversal via Drive-Relative Linkpath
- `GHSA-qffp-2rhf-9h96` — Hardlink Path Traversal via Drive-Relative Linkpath

Reach: `tar.extract()` is called from `js/src/download.ts:extractTar` to unpack the Chromium tarball under `~/.cloakbrowser/`. The existing entry `filter()` rejects absolute paths and `..` in the `name` field, but does not inspect the **linkpath** of a symlink/hardlink entry. A malicious Chromium mirror (or a downgrade-attack on the SHA256SUMS lookup, which warns-and-skips when missing) could ship a tarball with a drive-relative linkpath that writes outside the cache directory on Windows.

The other 6 patched CVEs are in **dev-only** transitive deps (`basic-ftp`, `ip-address`, `postcss`) so they do not ship to consumers of `cloakbrowser`, but they still affect anyone who runs `npm install` against this repo (CI, contributors).

## Location

`js/package-lock.json`

## Fix

| package      | before  | after   | CVEs                                   |
|--------------|---------|---------|----------------------------------------|
| tar          | 7.5.9   | 7.5.15  | GHSA-9ppj-qmqm-q256, GHSA-qffp-2rhf-9h96 (both HIGH, runtime) |
| basic-ftp    | 5.2.0   | 5.3.1   | GHSA-6v7q-wjvx-w8wg, -chqc-8p9q-pq6q, -rp42-5vxx-qpwr, -rpmf-866q-6p89 (4 × HIGH, dev) |
| ip-address   | 10.1.0  | 10.2.0  | GHSA-v2v4-37r5-5v8g (MOD, dev)         |
| postcss      | 8.5.6   | 8.5.14  | GHSA-qx2v-qp2m-jg93 (MOD, dev)         |

### Lockfile metadata side-effects

`npm audit fix --package-lock-only` also regenerated the lockfile metadata to match the current `package.json` (which had drifted ahead of the committed lockfile):

- top-level `name@version` (0.3.23 → 0.3.28)
- `devDependencies.playwright-core` range (^1.40.0 → ^1.53.0)
- `peerDependencies.playwright-core` range (>=1.40.0 → >=1.53.0)
- `peerDependencies.socks-proxy-agent` range (>=8.0.0 → >=10.0.0)

These are not editorial changes — they're npm bringing the lockfile back in sync with what `package.json` already declares. Happy to split them out into a separate cleanup PR if you'd prefer to keep this one pure-CVE.

## Detected by

Aeon + osv-scanner.
- Severity: high (runtime tar) / high+moderate (dev deps)
- 8 of 14 CVEs from the initial scan; remaining 6 (esbuild, tar-fs ×3, vite, ws) need a `puppeteer-core` 21 → 24 + `vitest` 1 → 3 bump, which is a breaking change.

## Verification

- `npm test` → **320 passed / 11 skipped / 0 failed** across 9 test files (including the tar-using download tests).
- `npm run typecheck` → clean.
- `osv-scanner` before this PR: 14 vulnerable advisories. After: 6 (all dev-only, all gated on the major bump above).

---

Filed by [Aeon](https://github.com/aaronjmars/aeon-aaron).

### DIFF
--- a/js/package-lock.json
+++ b/js/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "cloakbrowser",
-  "version": "0.3.23",
+  "version": "0.3.28",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "cloakbrowser",
-      "version": "0.3.23",
+      "version": "0.3.28",
       "license": "MIT",
       "dependencies": {
         "tar": "^7.0.0"
@@ -17,7 +17,7 @@
       "devDependencies": {
         "@types/node": "^20.10.0",
         "mmdb-lib": "^3.0.2",
-        "playwright-core": "^1.40.0",
+        "playwright-core": "^1.53.0",
         "puppeteer-core": "^21.0.0",
         "socks-proxy-agent": "^10.0.0",
         "typescript": "^5.3.0",
@@ -28,9 +28,9 @@
       },
       "peerDependencies": {
         "mmdb-lib": ">=2.0.0",
-        "playwright-core": ">=1.40.0",
+        "playwright-core": ">=1.53.0",
         "puppeteer-core": ">=21.0.0",
-        "socks-proxy-agent": ">=8.0.0"
+        "socks-proxy-agent": ">=10.0.0"
       },
       "peerDependenciesMeta": {
         "mmdb-lib": {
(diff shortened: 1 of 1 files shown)


Cloak-HQ:
Thanks for the thorough writeup. Merged.

The tar bump is the important one here — our extractTar filter checks entry names but not linkpaths, so this closes a real window on Windows. Good catch on scoping the remaining 6 to a separate major-bump PR.

PR #226, opened 2026-05-12
Add support for existing Playwright instances

SayanDey322:
This PR adds support for passing an existing Playwright instance and browser type into launch() instead of always creating a new internal Playwright session.

Changes:
- Added optional playwright parameter
- Added optional browser_type parameter
- Preserved backward compatibility
- Updated cleanup logic to avoid stopping external Playwright instances

### DIFF
--- a/cloakbrowser/browser.py
+++ b/cloakbrowser/browser.py
@@ -53,6 +53,8 @@ class ProxySettings(_ProxySettingsRequired, total=False):
 
 
 def launch(
+    playwright=None,
+    browser_type=None,    
     headless: bool = True,
     proxy: str | ProxySettings | None = None,
     args: list[str] | None = None,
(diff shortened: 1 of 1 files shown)


Cloak-HQ:
Thanks for the contribution! The idea of reusing an existing Playwright instance makes sense in theory.

However, after looking at this more closely, we don't think this is the right addition for CloakBrowser. Each `launch()` call intentionally owns its full lifecycle — creating and tearing down its own Playwright process. This keeps cleanup predictable and avoids shared-state issues (e.g. one `browser.close()` accidentally stopping a Playwright instance that other browsers still depend on).

Sharing a Playwright server process also doesn't save much in practice. The server itself is lightweight. It's the Chromium processes that use real resources, and those are separate regardless.

Adding this properly would mean touching all 6 Python entry points + 3 JS entry points, plus tests and docs, for a use case we haven't seen demand for.

If you have a specific integration scenario that requires this, we'd love to hear about it — it might point us toward a better API shape. But as-is, we're going to pass on this one.

Cloak-HQ:
Closing per the review above. Each launch() call intentionally owns its full lifecycle. Thanks for the idea though.

PR #227, opened 2026-05-12
Harden archive extraction and add a doctor CLI

HariVeeraManiKumarVallu:
## Summary
This PR improves CloakBrowser’s reliability on Windows and adds a lightweight diagnostics command for local setup checks.

## Changes
- Hardened archive extraction to use real path ancestry checks instead of string-prefix comparisons
- Fixed Windows executable detection in the binary cache path
- Added `cloakbrowser doctor` for environment, cache, binary, and optional dependency diagnostics
- Added regression tests for archive traversal, Windows paths, and CLI diagnostics
- Made a few path assertions platform-neutral so the test suite behaves correctly on Windows

## Verification
- `python -m pytest tests\test_extract.py tests\test_update.py tests\test_proxy.py tests\test_build_args.py tests\test_geoip.py tests\test_cli.py`
- `python -m pytest tests\test_config.py tests\test_cloakserve.py`

## Notes
This change is focused on safety and developer experience. It does not alter the browser launch API.

### DIFF
--- a/cloakbrowser/__main__.py
+++ b/cloakbrowser/__main__.py
@@ -10,7 +10,9 @@ Usage:
 from __future__ import annotations
 
 import argparse
+import importlib.util
 import logging
+import platform
 import sys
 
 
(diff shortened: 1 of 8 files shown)


Cloak-HQ (commented):
Thanks for the PR. The path traversal fix is solid, that string-prefix check is a known footgun and this is the right approach.

A few things before merge:

**Drop the `args.ts` change.** It guards `process` with optional chaining on line 8 (`nodeProcess?.env?.DEBUG`) but not on line 30 (`nodeProcess.platform`), so it crashes if `process` is actually undefined. And only `args.ts` gets this treatment while 5 other files in `js/src/` use `process` directly. This is incomplete and introduces a bug. Please revert and handle it in a separate PR if there's a use case.

**Use the built-in instead of the helper.** Python 3.9+ has `Path.is_relative_to()` and we require `>=3.9`. The `_path_is_relative_to()` function can be replaced with:
```python
if not member_path.is_relative_to(dest_resolved):
```

**Merge `doctor` into `info`.** The only new things `doctor` adds over `info` are Python version, OS, and module availability. That doesn't justify a separate command. Could you add those fields to `info` instead?

**Fix the test mock.** `test_doctor_prints_environment_without_downloading` mocks `ensure_binary`, but the function calls `binary_info()`, which never triggers a download. The mock is a no-op. Either remove it or mock something the function actually calls.

**Minor:**
- Remove `# type: ignore[import-not-found]` on the pytest import in `test_extract.py`. No other test file has this.

The extraction fix and its tests are good. The Windows `_is_executable` improvement and platform-neutral test assertions are welcome too. Just needs these cleanups.

Let me know if you have questions on any of these.

PR #228, opened 2026-05-12
feat(humanize): add Playwright-style actionability checks

Cloak-HQ:
## Summary

- Adds Playwright-style actionability checks (attached, visible, stable, enabled, editable, pointer-events) to all humanized interaction methods
- Before: hidden elements crashed with generic `RuntimeError`, disabled buttons were clicked, overlays were clicked through
- After: descriptive errors (`ElementNotVisibleError`, `ElementNotEnabledError`, `ElementNotReceivingEventsError`, etc.) with retry+backoff, `force=True` escape hatch
- Applies to all 13 page-level methods (click, dblclick, hover, type, fill, check, uncheck, selectOption, press, pressSequentially, tap, clear, dragTo) in both sync and async paths, plus Python and JS/TS

### New files
- `cloakbrowser/human/actionability.py` — error hierarchy, check-set constants, sync functions
- `cloakbrowser/human/actionability_async.py` — async mirror
- `js/src/human/actionability.ts` — TypeScript implementation

### Key design decisions
1. **Shared deadline** — one timeout budget across all steps (checks + scroll + stable + pointer-events)
2. **`_skip_checks`** for chained methods — outer method runs checks, inner click skips pre-scroll checks
3. **`force=True`** — skips all checks, matches Playwright semantics
4. **Stability check optimization** — only runs if `scroll_to_element` actually scrolled
5. **Pointer-events check uses `locator.evaluate()`** — works with all Playwright selector types (text=, role=, XPath, CSS)

## Test plan

- [x] Unit tests: 60 Python passed, 71 JS passed
- [x] Docker integration test (`debug/test_locator_retry.py`): 8/8 pass
- [x] Full manual test suite (`debug/test_actionability_full.py`): 34/34 pass (sync + async)
  - Basic click, delayed auto-retry, hidden/disabled/overlay errors, force=True, type, fill, check, uncheck, select_option, hover, dblclick, locator API, typing speed
- [x] 3 rounds of multi-model code review (Opus + Pro + Codex), all CRITICAL/HIGH/MEDIUM issues resolved

@evelaa123 — you built the humanize layer, would love your review on how actionability checks integrate with the existing mouse/keyboard/scroll pipeline.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ Open [http://localhost:8080](http://localhost:8080). Create a profile. Click **L
 
 ---
 
-## Latest: v0.3.26 (Chromium 146.0.7680.177.4)
+## Latest: v0.3.28 (Chromium 146.0.7680.177.4)
 
 - **`launch_context_async()`** — async counterpart to `launch_context()`. Forwards kwargs to `browser.new_context()` for `storage_state`, `permissions`, `extra_http_headers` without a persistent profile folder.
 - **JS `contextOptions` escape hatch** — forward arbitrary options (including `storageState`) to Playwright's `newContext()` from `launchContext()` / `launchPersistentContext()`.
--- a/cloakbrowser/human/__init__.py
+++ b/cloakbrowser/human/__init__.py
@@ -1070,7 +1070,8 @@ def patch_page(page: Any, cfg: HumanConfig, cursor: _CursorState) -> None:
             ensure_actionable(page, selector, CHECKS_FOCUS, timeout=_remaining_ms(), force=force)
         _human_hover(selector, _skip_checks=True, timeout=_remaining_ms(), force=force, human_config=kwargs.get("human_config"))
         sleep_ms(rand(100, 300))
-        return originals.select_option(selector, value, **kwargs)
+        pw_kwargs = {k: v for k, v in kwargs.items() if k not in ("human_config", "force")}
+        return originals.select_option(selector, value, **pw_kwargs)
 
     def _human_press(selector: str, key: str, **kwargs: Any) -> None:
         force = kwargs.get("force", False)
@@ -1984,7 +1985,8 @@ def patch_page_async(page: Any, cfg: HumanConfig, cursor: _CursorState) -> None:
             await async_ensure_actionable(page, selector, CHECKS_FOCUS, timeout=_remaining_ms(), force=force)
         await _human_hover(selector, _skip_checks=True, timeout=_remaining_ms(), force=force, human_config=kwargs.get("human_config"))
         await async_sleep_ms(rand(100, 300))
-        return await originals.select_option(selector, value, **kwargs)
+        pw_kwargs = {k: v for k, v in kwargs.items() if k not in ("human_config", "force")}
+        return await originals.select_option(selector, value, **pw_kwargs)
 
     async def _human_mouse_move(x: float, y: float, **kwargs: Any) -> None:
         await _ensure_cursor_init()
--- a/cloakbrowser/human/__init__.py
+++ b/cloakbrowser/human/__init__.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 import json
 import logging
 import sys
+import time
 from typing import Any, Optional
 
 from .config import HumanConfig, HumanPreset, resolve_config, merge_config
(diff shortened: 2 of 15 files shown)


evelaa123:
Thank's for mention! I will be happy to read in detail and tell you my opinion within 12 hours!

evelaa123:
Thanks for the work on this — the Playwright-style actionability layer is a solid addition. The error hierarchy mirrors upstream nicely and the `[100, 250, 500, 1000]` backoff ladder matches Playwright exactly. Verified locally against the `pr-228` branch; found a few issues worth addressing before merge.

### 🔴 BUG 1 — `human_config` is dropped in 8 wrappers (sync + async)

In `cloakbrowser/human/__init__.py`, the inner click/hover call inside `_human_check`, `_human_uncheck`, `_human_select_option`, `_human_press` (both sync and async) does **not** forward `human_config`:

```python
# current — sync _human_check / _human_uncheck / _human_press
_human_click(selector, _skip_checks=True, timeout=_remaining_ms(), force=force)

# current — sync _human_select_option
_human_hover(selector, _skip_checks=True, timeout=_remaining_ms(), force=force)
```

Compare to `_human_type` and `_human_fill`, which forward it correctly:

```python
_human_click(selector, _skip_checks=True, timeout=_remaining_ms(),
             force=force, human_config=kwargs.get("human_config"))
```

The same omission appears in all four async counterparts. Net effect: `page.check("#x", human_config={"key_hold": (999, 999)})` runs the inner click with default timing, silently ignoring the per-call override.

**Fix:** add `human_config=kwargs.get("human_config")` to all 8 inner-call sites (4 sync + 4 async).

### 🟡 BUG 2 — `check_pointer_events_handle` interpolates coords via f-string

The selector variant in `actionability.py` passes coordinates as evaluate args:

```python
# selector variant — static JS, coords passed as second arg
_POINTER_EVENTS_LOCATOR_JS = """(expected, coords) => {
    const target = document.elementFromPoint(coords.x, coords.y);
    ...
}"""
loc.evaluate(_POINTER_EVENTS_LOCATOR_JS, coords)
```

The handle variant interpolates coords into the JS source via f-string:

```python
js = f"""(expected) => {{
    const target = document.elementFromPoint({x}, {y});
    ...
}}"""
result = el.evaluate(js)
```

Not a security issue (no user input reaches this) and not a correctness issue with current callers (coords are always finite floats). Flagging it as a consistency nit — the two variants should follow the same pattern. Same in `actionability_async.py`.

**Fix:** mirror the selector pattern — use a static JS template and pass `{"x": x, "y": y}` as the second arg to `el.evaluate(js, {"x": x, "y": y})`.

### 🟡 MAINTAINABILITY — deadline boilerplate is duplicated ~18 times

`grep -c "time.monotonic()" cloakbrowser/human/__init__.py` returns 36, which corresponds to ~18 copies of the 4-line block (each block uses `time.monotonic()` twice). The identical block is repeated across 9 sync wrappers, their 9 async twins, and the handle-level methods:

```python
timeout = kwargs.get("timeout", 30000)
force = kwargs.get("force", False)
deadline = time.monotonic() + timeout / 1000.0
def _remaining_ms():
    return max(0, (deadline - time.monotonic()) * 1000)
```

Any future tweak (logging, default timeout change, sub-action budget split) will almost certainly miss at least one copy.

Suggested helper:

```python
def _make_deadline(kwargs, default_timeout=30000):
    timeout = kwargs.get("timeout", default_timeout)
    force = kwargs.get("force", False)
    deadline = time.monotonic() + timeout / 1000.0
    return force, lambda: max(0, (deadline - time.monotonic()) * 1000)
```

### 🟢 API change worth documenting

`scroll_to_element` / `human_scroll_into_view` now return a 4-tuple `(box, cx, cy, did_scroll)` instead of the previous 3-tuple. This is a breaking change for any external code that unpacks the return value. Worth calling out explicitly in the PR description and CHANGELOG.

### Suggested patch summary

1. Add `human_config=kwargs.get("human_config")` to the 8 inner-call sites (sync + async `_human_check`, `_human_uncheck`, `_human_select_option`, `_human_press`). Recommend two regression tests asserting `human_config={"key_hold": (999, 999)}` propagates through `page.check()` and `page.uncheck()`.
2. Convert `check_pointer_events_handle` (sync + async) to pass coords as evaluate args, matching the selector variant.
3. Optional but recommended: extract a `_make_deadline(kwargs)` helper and use it in all ~18 wrappers.
4. Note the `scroll_to_element` return-shape change in the PR description and CHANGELOG.

Cloak-HQ:
Thanks for the thorough review! All four items addressed:

**BUG 1 — `human_config` forwarding** (f4f6a3f)
Added `human_config=kwargs.get("human_config")` to all 8 inner-call sites (sync + async check/uncheck/select_option/press) plus 6 JS equivalents. Also found that `originals.select_option(**kwargs)` was forwarding our custom kwargs (`human_config`, `force`) to Playwright, which rejects unknown kwargs. Stripped them before forwarding (749ed56).

**BUG 2 — `check_pointer_events_handle` f-string** (f4f6a3f)
Converted to `el.evaluate(js, coords)` pattern matching the selector variant. Applied to Python sync, async, and JS.

**Maintainability — deadline boilerplate**
Good suggestion. Deferring to a follow-up. The current shape works and extracting a helper is a pure refactor.

**API change — scroll return shape**
Internal API only (`scroll_to_element` / `human_scroll_into_view` aren't exported), but worth noting for anyone extending the humanize layer.

Verified with 42 Docker integration tests (23 sync + 19 async) including dedicated human_config propagation tests for check/uncheck/select_option/press. CI green.

Issue #160, opened 2026-04-17
Performance decrease with multiple contexts


curvah:
Hi guys, when running N (from my tests 4 is enough to replicate) threads using a single shared Browser instance with one BrowserContext per thread, the turnstile challenge take significantly longer to resolve (or stall indefinitely) compared to spawning a dedicated Browser instance per task.

This is how I was creating the browser and the context (pretty straightforward):
```python
browser = await launch_async(
        headless=False,
        geoip=False,
        humanize=True,
)
context = await browser.new_context(proxy={"server": "...", "username": "...", "password": "..."})
```

I tested the same script but creating one browser per task instead of using contexts and the results where as expected, each task managed to solve the challenge without any problem, but I had to use geoip=True (that I can't when using contexts).

Using VNC to see why its taking much to solve seems like that cloudflare is detecting something or something is not going as it should be (the "Stuck?" warning is a sign that is taking too much to load the challenge):
<img width="1247" height="820" alt="Image" src="https://github.com/user-attachments/assets/7d24cb04-6104-4089-9f16-128f0fa2c0cc" />

I am running this on docker with xvfb and the cloakbrowser version is 146.0.7680.177.3

Is this expected behavior, or does it point to a bug in context isolation?
Im always open to test more if needed, thanks in advance for any help  :)

Cloak-HQ:
Thanks for the report.

We tested your scenario on our end and found the root cause isn't context isolation — it's a geoip limitation you already noticed.

**What's happening**

`timezone` and `locale` are process-level binary flags, set once when the Browser launches. `geoip=True` only resolves them if the proxy is passed to `launch_async()` — not when the proxy is set on `new_context()`. We confirmed this:

- `launch_async(geoip=True) + new_context(proxy=DE)` → `timezone=UTC`, `locale=en-US`
- `launch_async(geoip=True, proxy=DE) + new_context()` → `timezone=Europe/Berlin`, `locale=de-DE`

So in your shared-Browser setup, every context runs with UTC + en-US regardless of each proxy's country. That geo mismatch is a strong signal for Turnstile's managed challenge → interactive mode → "Stuck?" stall. One Browser per task works because each launch can resolve geoip properly.

We were not able to reproduce the stall on the official Turnstile demo (`demo.turnstile.workers.dev`) — 4 concurrent contexts on a shared Browser solved in 3-5s each. So the concurrency itself is fine; the issue is the geo mismatch.

**Workarounds**

[...]

PR #229, opened 2026-05-12
fix(geoip): warn when proxy is set per-context (geoip resolves at launch only)

mvanhorn:
## Summary

When `geoip=True` is set on `launch_async()` but a later `new_context()` gets its own `proxy=`, the per-context proxy doesn't trigger the geoip resolution path — `timezone` and `locale` stay set to whatever the launch resolved. This now warns once with a pointer at the fix shape, instead of silently looking detectable.

The warning is conservative: it doesn't promise behavior the binary patches don't deliver. The deeper "per-context timezone/locale spoofing" change is binary-side and out of scope for this PR.

## What changed

- `cloakbrowser/browser.py:44` — patches sync and async `Browser.new_context()` to detect the misconfiguration (geoip enabled at Browser level + proxy passed per-context) and emit a one-time `warnings.warn(..., UserWarning)`.
- `js/src/playwright.ts:47` — matching JS `browser.newContext()` warning.
- `README.md` and `js/README.md` — "GeoIP + multiple contexts" subsection documenting the limitation and how to avoid it.
- `tests/test_context_geoip_warning.py` + `js/tests/launch.test.ts` — warn-once, quiet-path (proxy on launch), quiet-path (no geoip), quiet-path (no secret).

## Notes for [#160](https://github.com/CloakHQ/CloakBrowser/issues/160)

The thread accepted the maintainer's diagnosis but raised concern that something else might be off. This PR ships the diagnosis as a warning + docs so users hit a clear signal. If the deeper investigation turns up a per-context resolution bug beyond geoip, that's a separate change.

## Testing

- Python: `pytest tests/test_context_geoip_warning.py tests/test_launch_context.py tests/test_proxy.py` — 79/79 passed
- `python3 -m py_compile cloakbrowser/browser.py tests/test_context_geoip_warning.py` — clean
- JS: `npm ci` blocked by sandbox network (`ENOTFOUND registry.npmjs.org`); JS test follows the existing shape, CI will exercise.

Refs #160

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,6 +284,14 @@ browser = launch(humanize=True, human_preset="careful")
 browser = launch(stealth_args=False, args=["--fingerprint=12345"])
 ```
 
+#### geoip warnings
+
+When you call `cloakbrowser.launch(geoip=True)` without `proxy=`, the library
+logs a warning. Geoip timezone and locale resolution only runs at launch with
+the launch-time proxy. Per-context proxies set later with
+`browser.new_context(proxy=...)` will not trigger geoip resolution. Pass
+`proxy=` to `launch()` or set explicit `timezone=` / `locale=` to silence it.
+
 Returns a standard Playwright `Browser` object. All Playwright methods work: `new_page()`, `new_context()`, `close()`, etc.
 
 ### `launch_async()`
--- a/README.md
+++ b/README.md
@@ -329,6 +329,14 @@ browser = launch(humanize=True, human_preset="careful")
 browser = launch(stealth_args=False, args=["--fingerprint=12345"])
 ```
 
+#### geoip warnings
+
+When you call `cloakbrowser.launch(geoip=True)` without `proxy=`, the library
+logs a warning. Geoip timezone and locale resolution only runs at launch with
+the launch-time proxy. Per-context proxies set later with
+`browser.new_context(proxy=...)` will not trigger geoip resolution. Pass
+`proxy=` to `launch()` or set explicit `timezone=` / `locale=` to silence it.
+
 Returns a standard Playwright `Browser` object. All Playwright methods work: `new_page()`, `new_context()`, `close()`, etc.
 
 ### `launch_async()`
--- a/cloakbrowser/browser.py
+++ b/cloakbrowser/browser.py
@@ -109,6 +109,12 @@ def launch(
 
     binary_path = ensure_binary()
     timezone, locale, exit_ip = maybe_resolve_geoip(geoip, proxy, timezone, locale)
+    if geoip and not proxy:
+        logger.warning(
+            "geoip=True without proxy= at launch - timezone/locale will default to "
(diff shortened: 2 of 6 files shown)


Cloak-HQ (commented):
Thanks for the PR and for documenting this limitation clearly.

The monkey-patching approach adds more complexity than needed for a warning. Two new functions (sync + async), duplicated `_has_context_proxy` helpers, and it still misses `browser.new_page(proxy=...)` which bypasses the patched `new_context` internally.

Simpler alternative: warn at launch time when `geoip=True` but no proxy is passed to launch. That's the actual misconfiguration moment from #160.

```python
# In launch() / launch_async(), after maybe_resolve_geoip:
if geoip and not proxy:
    logger.warning(
        "geoip=True without proxy= at launch — timezone/locale will default to "
        "UTC/en-US. Per-context proxies don't trigger geoip resolution. Pass "
        "proxy= to launch() or set explicit timezone=/locale=."
    )
```

Same for JS. No new functions, no monkey-patching, covers all code paths including `new_page`. The README/docs additions are good as-is.

There's an edge case this doesn't cover: `launch(proxy=A, geoip=True)` then `new_context(proxy=B)` where B is a different region. But that's already documented by your README addition, so users have the info they need.

mvanhorn:
Thanks - rewrote per the suggestion. Pushed 38197a2.

The branch was force-pushed: I reset to current main, dropped the monkey-patching, and reimplemented as a launch-time `if geoip and not proxy: warn(...)` inline in each public entry point (`launch`, `launch_async`, `launch_persistent_context*`, `launch_context*`). The `launch_context*` wrappers resolve geoip eagerly and call the lower-level `launch()` without forwarding `geoip=`, so the warning fires exactly once per user-facing call - no double-warning bug. Same pattern in `js/src/playwright.ts` across `buildLaunchOptions`, `launchContext`, and `launchPersistentContext`.

Covers `browser.new_page(proxy=...)` and any other post-launch context routing because the warning fires at launch regardless of what happens after. Tests in `tests/test_geoip_launch_warning.py` (pytest caplog) and `js/tests/launch.test.ts` (vitest console.warn spy) cover the fires-when-no-proxy and silent-when-proxy-set cases. README + `js/README.md` get short geoip-warnings subsections.

Let me know if you want anything tightened.

mvanhorn:
Good call - switched to the launch-time warning approach you suggested and dropped the monkey-patching entirely. When geoip=True but no proxy is passed to launch, it now warns (your exact message) in launch()/launch_async() on the Python side and in buildLaunchOptions() on the JS side, so it covers every path including new_page. Kept the README limitation note. Rebased onto main.

Verified: the new warning tests pass (fires when geoip + no proxy, stays quiet when a launch proxy is set, both sync and async). I couldn't run the JS Vitest locally (no js/node_modules here), but the JS change is the same guarded console.warn.

Issue #232, opened 2026-05-13
Direct CDP ws connection url


400ochkov:
Hello, is it possible to get direct CDP ws connection url? I have my own CDP SDK and want to use it with CloakBrowser. I tried CloakBrowser Manager but it works via docker and create connection url via localhost:8080, so I can't to connect to it. May be there is some convinience way to get direct CDP connection url?

400ochkov:
I used launch_persistent_context with args=["--remote-debugging-port=9222"] and then get ws url from http://localhost:9222/json/version but this way not convinience and moreover after connecting to ws endpoint browser stop loading anything, like no have internet (but everything is ok with internet), even when stop ws connection nothing loaded.
Is it bug?

Cloak-HQ:
Thanks for the report.

**Not a bug** — this is expected behavior when mixing Playwright's managed connection with an external CDP client.

**Why it happens**: When you call `launch_persistent_context`, Playwright connects to Chrome via `--remote-debugging-pipe` (stdio). Your `--remote-debugging-port=9222` opens a second CDP endpoint alongside it. Chrome accepts both, but when an external client connects to port 9222, Playwright's internal session management gets disrupted — the browser loses the ability to load pages, and the state doesn't recover even after disconnecting.

**If you need CDP access from within your script**, use Playwright's built-in CDP session (no second connection needed):

```python
ctx = launch_persistent_context("./my-profile", headless=False)
page = ctx.new_page()
page.goto("https://example.com")

cdp = ctx.new_cdp_session(page)
result = cdp.send("Runtime.evaluate", {"expression": "navigator.userAgent"})
```

**If you need external CDP access** (separate process, DevTools frontend, another tool), use `cloakserve` — it launches Chrome directly without Playwright, so there's no session conflict:

```bash
cloakserve --port=9222
```

Then connect from any client:

```python
from playwright.sync_api import sync_playwright

pw = sync_playwright().start()
browser = pw.chromium.connect_over_cdp("http://localhost:9222?fingerprint=12345")
page = browser.contexts[0].pages[0]
```

`cloakserve` supports persistent data dirs (`--data-dir=./my-profile`), per-connection fingerprint seeds, timezone/locale, and proxy — all via query params.

Related: #126 (humanize over CDP).

Let us know which approach fits your use case, or if you need something different.

AsteroidCosmic:
[...]

PR #234, opened 2026-05-13
fix: rewrite cloakserve CDP WebSocket URLs

honor2030:
## Summary
- rewrite `cloakserve` discovery `webSocketDebuggerUrl` values through the public `cloakserve` host instead of the internal Chrome CDP host
- honor `X-Forwarded-Host` and `X-Forwarded-Proto` so reverse-proxy/TLS deployments return reachable `ws`/`wss` endpoints
- document how direct CDP SDK users can fetch `/json/version` and use the rewritten WebSocket URL

Addresses #232

## Test Plan
- `python3 -m py_compile bin/cloakserve`
- `git diff --check origin/main...HEAD`
- `python3 -m pytest tests/test_cloakserve.py -q` → `61 passed`
- `python -m pytest -q` currently fails in this local environment because `playwright` is not installed (`ModuleNotFoundError: No module named 'playwright'`); the focused cloakserve suite above passes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -808,6 +808,26 @@ print(page.title())
 browser.close()
 ```
 
+If your framework needs a direct WebSocket endpoint, fetch Chrome's discovery document and use the rewritten `webSocketDebuggerUrl`. The URL points back through `cloakserve` so the CDP proxy can keep per-seed routing intact:
+
+```bash
+curl http://localhost:9222/json/version | jq -r .webSocketDebuggerUrl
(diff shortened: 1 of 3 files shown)


Cloak-HQ:
Thanks for the contribution! Merged — the `X-Forwarded-Host` / `X-Forwarded-Proto` support is a clean addition for reverse proxy setups. Let us know if you run into any issues.

PR #235, opened 2026-05-13
fix: warn on unauthenticated public cloakserve binds

honor2030:
## Summary
- Add a cloakserve diagnostic warning when the service binds to all interfaces without authentication.
- Surface the warning in startup logs and `GET /` diagnostics so unsafe CDP exposure is visible.
- Document the warning next to the existing CDP security guidance.

## Test Plan
- RED: `python -m pytest tests/test_cloakserve.py::TestBindSecurityWarnings -q` failed before implementation with `AttributeError: module 'cloakserve' has no attribute '_bind_security_warnings'`.
- GREEN: `python -m pytest tests/test_cloakserve.py -q` → `56 passed`.
- `python -m py_compile bin/cloakserve`
- `git diff --check`

## Notes
- Checked PR #227 for overlap. It touches `tests/test_cloakserve.py` for a separate `_default_data_dir` assertion plus archive/doctor CLI files; this PR keeps the scope to cloakserve public-bind diagnostics.
- Broader local `python -m pytest -q` is blocked by the local environment missing the optional `playwright` dependency (`ModuleNotFoundError: No module named 'playwright'`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -829,6 +829,8 @@ docker stop cloak && docker rm cloak
 > **Security:** CDP gives full control over the browser (execute JS, read pages, access files).
 > The examples bind to `127.0.0.1` so only your machine can connect. Never expose port 9222
 > to the public internet without additional authentication.
+> In containers, `cloakserve` listens on `0.0.0.0` internally so Docker port publishing works;
+> if the service is reachable on all interfaces, startup logs and `GET /` diagnostics include a warning.
 
 ### Docker Compose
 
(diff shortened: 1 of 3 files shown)


honor2030:
Scope note to keep the cloakserve queue easier to review.

This PR is intentionally warning-only: it does not add authentication or change access control. It just surfaces the existing risk in startup logs and the root diagnostics when `cloakserve` is listening on a public bind.

I saw that #248 touches the same cloakserve surface with a broader `--auth-token` / explicit-host hardening change. If #248 lands and maintainers feel its warning path covers this diagnostic use case, I’m happy to rebase or close this PR to avoid duplicate review. Otherwise, this can stay as a small diagnostic-only fallback.

Cloak-HQ:
Closing in favor of #248 which includes this warning plus opt-in `--auth-token` support. Thanks for the contribution!

PR #240, opened 2026-05-14
fix: guard cloakserve WebSocket origins

### LINKED ISSUE #239: Security: Untrusted browser origins can access local `cloakserve` CDP, leading to unauthorized browser control and local file disclosure
## Description

`cloakserve` accepts WebSocket connections from untrusted browser origins and transparently proxies them to the local Chromium DevTools socket on `127.0.0.1`, without validating the request `Origin` or requiring authentication.

In practice, this allows JavaScript running in a real public webpage to obtain unauthorized CDP access to a local CloakBrowser instance. Once CDP access is obtained, the remote page can drive the browser, inspect page contents, execute JavaScript, take screenshots, navigate to `file://` URLs, and read local file contents through DOM inspection.

This is not only a non-browser client issue. I reproduced the core behavior with JavaScript running inside a real public webpage in default Chromium-based browsers. The same public webpage could connect successfully to local `cloakserve`, while the same browser-origin WebSocket connection sent directly to raw Chrome DevTools was rejected with HTTP 403.

The security boundary issue is that `cloakserve` re-exposes a localhost-protected DevTools interface to browser-origin JavaScript, while upstream Chrome correctly blocks the same browser-origin connection.

I would like to request that the maintainers assess this as a security issue and, if confirmed, consider GHSA/CVE assignment.

## Expected behavior

`cloakserve` should reject untrusted browser-origin WebSocket upgrade requests before proxying them to the local DevTools endpoint.

At minimum, `cloakserve` should either:

- validate `Origin` against a strict allowlist;
- require an explicit authentication token for `/json/*` and `/devtools/*`;
- reject browser-origin WebSocket connections by default unless explicitly enabled.

## Actual behavior

A remote public webpage can establish a browser-origin WebSocket connection to local `cloakserve` and issue CDP commands against the local browser.

As a control, the same browser-origin WebSocket connection sent directly to raw Chrome DevTools is rejected with HTTP 403.

## Impact

A malicious webpage opened by the user can obtain unauthorized access to the local CloakBrowser CDP control plane if `cloakserve` is running locally and reachable from the browser.

Confirmed impact includes:

- unauthorized browser-origin access to the local CloakBrowser CDP control plane;
- control over the managed browser through CDP commands;
- access to page contents, screenshots, runtime data, and automation state inside the managed browser instance;
- local file disclosure by navigating the controlled browser to `file://` URLs and reading the resulting DOM;
- bypass of Chromium’s upstream localhost DevTools protection, because raw Chrome rejects the same browser-origin connection.

The local file disclosure impact is a consequence of unauthorized CDP control. The readable scope is limited to files accessible to the managed browser process and renderable through Chromium’s `file://` handling.

## Severity / CVE request

Suggested severity: High.

A conservative CVSS vector may be:

`CVSS:3.1/AV:N/AC:L/PR:N/UI:R/S:C/C:H/I:H/A:L`

I understand the final severity may depend on the maintainers’ threat model for local `cloakserve` and browser-origin localhost access. However, I believe this is CVE-worthy because the affected component is the core `cloakserve` CDP proxy, and the issue allows a remote webpage to cross the expected local DevTools security boundary and obtain unauthorized CDP control.

This is not simply a case of exposing the debugging port to the public Internet. The reproduced scenario uses a locally running `cloakserve` instance and a public webpage opened by the victim. The issue is that the proxy layer does not enforce the browser-origin protections that raw Chromium DevTools applies.

## CloakBrowser version

Confirmed on:

- `0.3.28`
- current HEAD: `ad4d946ca6c5a1490223818e21e8241910deed8b`

This appears likely to affect the multiplexer-based `cloakserve` design introduced around `0.3.20`, but I have directly verified the issue only on `0.3.28` / current HEAD.

## Wrapper

Python

## Environment

- Windows 11 x64
- Bare metal, not Docker
- Local `cloakserve` listening on `127.0.0.1:9555`
- Backing Chromium DevTools socket on `127.0.0.1:5100`
- Reproduced from a real public remote HTTP origin in default Chromium-based browsers:
  - Chrome `148.0.7778.97`
  - Edge `148.0.3967.54`

## Launch options

```bash
cloakserve --port 9555
```

## Tested with a different IP or proxy?

No proxy was required. The issue was reproduced from a public remote origin.

## Works outside Docker / on host machine?

Yes.

## Steps to reproduce

1. Start `cloakserve` on the host machine:

```bash
cloakserve --port 9555
```

2. Confirm that the local DevTools path exists. For example, locally query:

```text
http://127.0.0.1:9555/json/version
```

and note the returned browser WebSocket path:

```text
ws://127.0.0.1:9555/devtools/browser/<GUID>
```

3. Host a simple HTML page on a public remote server, for example:

```text
http://PUBLIC_HOST:9555
```

The page should:

- call `new WebSocket("ws://127.0.0.1:9555/devtools/browser/<GUID>")`;
- send a basic CDP command such as `Browser.getVersion`;
- display success or failure in the page body.

4. Open that public page in a default local Chrome or Edge browser.

5. Observe that the WebSocket connection opens successfully and that CDP replies are received from the local CloakBrowser instance.

6. Extend the same public page to send CDP commands that create a target, navigate to a safe temporary local file such as:

```text
file:///C:/.../cloakbrowser-public-origin-read.txt
```

and read back:

```javascript
document.body.innerText
```

7. Observe that the unique marker string from the local file is displayed by the public page.

8. As a control test, change only the WebSocket target to raw Chrome DevTools:

```text
ws://127.0.0.1:5100/devtools/browser/<GUID>
```

and reload the same public page.

9. Observe that the direct connection to raw Chrome fails with HTTP 403, showing that Chrome itself rejects the untrusted browser origin while `cloakserve` accepts and proxies it.

## Observed results

Successful public-origin WebSocket to local `cloakserve`:

```text
origin=http://PUBLIC_HOST:9555
event=open
event=message data={"id":1,"result":{"protocolVersion":"1.3","product":"Chrome/146.0.7680.177",...}}
result=success
```

Successful local file disclosure from the public page:

```text
file-body=PUBLIC-ORIGIN-LOCAL-FILE-READ-POC-24680
result=success
```

Control test against raw Chrome DevTools:

Browser console:

```text
WebSocket connection to 'ws://127.0.0.1:5100/devtools/browser/<GUID>' failed: Error during WebSocket handshake: Unexpected response code: 403
```

Chrome log:

```text
Rejected an incoming WebSocket connection from the http://PUBLIC_HOST:9555 origin...
```

## Relevant code paths

The relevant code paths appear to be:

- `bin/cloakserve` lines `530-563`: accepts WebSocket upgrades and proxies them to `ws://127.0.0.1:<cdp_port>/devtools/...`
- `bin/cloakserve` lines `490-492`: establishes the backend CDP connection without enforcing browser-origin restrictions
- `bin/cloakserve` lines `421-433`: rewrites `webSocketDebuggerUrl` using the incoming `Host` header
- `bin/cloakserve` lines `462-474`: rewrites `/json/list` WebSocket URLs using the incoming `Host` header
- `bin/cloakserve` lines `373-392`: exposes unauthenticated runtime metadata at `/`

The missing `Origin` validation on the WebSocket proxy appears to be the primary issue. The `Host` header reflection in returned WebSocket URLs is related and may make origin/host confusion easier in some deployment or rebinding scenarios, but the core reproduced behavior is that a public browser origin can directly connect to local `cloakserve` and issue CDP commands.

## Why this is different from exposing CDP publicly

This report does not rely on exposing the CDP port to the public Internet.

The reproduced scenario is:

1. the victim runs `cloakserve` locally;
2. the victim opens a public remote webpage;
3. JavaScript in that public webpage connects to local `cloakserve`;
4. `cloakserve` accepts the browser-origin WebSocket and proxies it to the local DevTools socket;
5. the webpage obtains CDP control.

Raw Chrome DevTools rejects the same browser-origin WebSocket connection with HTTP 403. Therefore, the issue is not that CDP is generally dangerous, but that `cloakserve` weakens the upstream DevTools local security boundary by accepting and proxying untrusted browser-origin WebSocket connections.

## Recommended fixes

Possible mitigations:

1. Validate `Origin` on all `/devtools/*` WebSocket upgrade requests.
2. Reject untrusted browser origins by default.
3. Require a random bearer token or capability token for `/json/*`, `/devtools/*`, and `/`.
4. Avoid trusting the incoming `Host` header when generating `webSocketDebuggerUrl`.
5. Add a strict localhost `Host` allowlist for HTTP and WebSocket requests.
6. Minimize or authenticate the `GET /` status endpoint.
7. Avoid returning raw proxy configuration or other sensitive runtime metadata from unauthenticated endpoints.
8. Add regression tests showing that a browser-origin WebSocket from an untrusted public origin is rejected.

## Additional notes

I can provide the PoC page and full logs privately if preferred.

I intentionally avoided posting a more complete browser-control/file-enumeration PoC publicly. The examples above use only a safe temporary file with a unique marker string.

If the maintainers confirm this issue, I would appreciate GHSA/CVE assignment or guidance on the project’s preferred CVE process.

honor2030:
## Summary
- Add a WebSocket Origin guard to `cloakserve` before proxying CDP WebSocket upgrades.
- Preserve CLI/CDP clients that omit `Origin` and Chrome DevTools origins.
- Reject untrusted browser origins, public `Origin == Host` cases, malformed Origin/Host authorities, userinfo, empty ports, comma-separated hosts, and path/query/fragment forms.
- Cover both default and seeded `/devtools/*` WebSocket handlers with regression tests.

Closes #239

## Test Plan
- RED: `python -m pytest tests/test_cloakserve.py::TestWebSocketOriginGuard::test_public_origin_matching_host_is_still_rejected -q` failed before the loopback-only Host restriction because `http://attacker.example:9555` matching `attacker.example:9555` was allowed.
- GREEN: `python -m pytest tests/test_cloakserve.py::TestWebSocketOriginGuard -q` → 30 passed.
- GREEN: `python -m pytest tests/test_cloakserve.py -q` → 82 passed.
- GREEN: `python -m pytest tests/test_cloakserve.py tests/test_build_args.py tests/test_config.py tests/test_update.py tests/test_backend.py tests/test_geoip.py tests/test_proxy.py tests/test_extract.py -q` → 273 passed.
- GREEN: `python -m py_compile bin/cloakserve tests/test_cloakserve.py`.
- GREEN: `git diff --check` and added-line static scans for secrets/shell/eval/pickle/sql.
- Full local suite attempted: `python -m pytest tests -q --tb=short` currently fails because this local environment does not have the required `playwright` module installed; the failures are outside this PR's diff scope.

## Notes
- The guard intentionally does not add authentication; it fails closed for browser-origin WebSocket CSRF while keeping existing non-browser CDP clients compatible.

### DIFF
--- a/bin/cloakserve
+++ b/bin/cloakserve
@@ -18,6 +18,7 @@ Client:
 from __future__ import annotations
 
 import asyncio
+import ipaddress
 import json
 import logging
 import os
@@ -29,7 +30,7 @@ import subprocess
 import sys
 import time
 from dataclasses import dataclass
-from urllib.parse import parse_qs
+from urllib.parse import parse_qs, urlparse
 
 from pathlib import Path
 
@@ -64,6 +65,91 @@ BASE_CDP_PORT = 5100
 
 SAFE_SEED_RE = re.compile(r"^[A-Za-z0-9_-]{1,128}$")
 RESERVED_SEEDS = {"__default__"}
+TRUSTED_WS_ORIGINS = {"devtools://devtools", "chrome-devtools://devtools"}
+
+
+def _host_port_from_netloc(netloc: str, default_port: int) -> tuple[str, int] | None:
+    """Return a normalized (host, port) pair for an Origin/Host netloc."""
+    if "," in netloc:
+        return None
+    try:
+        parsed = urlparse(f"//{netloc.strip()}")
+        authority = parsed.netloc.rsplit("@", 1)[-1]
+        if (
+            not parsed.hostname
+            or parsed.username is not None
+            or parsed.password is not None
+            or authority.endswith(":")
+            or parsed.path
+            or parsed.params
+            or parsed.query
+            or parsed.fragment
+        ):
+            return None
+        return (parsed.hostname.lower(), parsed.port if parsed.port is not None else default_port)
+    except ValueError:
+        return None
+
+
+def _is_loopback_host(hostname: str) -> bool:
+    """Return True for localhost and loopback IP literals."""
+    hostname = hostname.strip("[]").rstrip(".").lower()
+    if hostname == "localhost":
+        return True
+    try:
+        return ipaddress.ip_address(hostname).is_loopback
+    except ValueError:
+        return False
+
+
+def _origin_is_allowed(
+    origin: str | None,
+    host: str | None,
+    request_scheme: str = "http",
+) -> bool:
+    """Return True when a WebSocket Origin is safe to proxy to local CDP."""
+    if origin is None:
+        # Playwright/Puppeteer and other non-browser CDP clients commonly omit
+        # Origin. Keep those clients working while rejecting browser-origin CSRF.
+        return True
+
+    origin = origin.strip()
+    if not origin or origin.lower() == "null":
+        return False
+    if origin in TRUSTED_WS_ORIGINS:
+        return True
+
+    try:
+        parsed = urlparse(origin)
+    except ValueError:
+        return False
+
+    if parsed.scheme not in ("http", "https"):
+        return False
+    if parsed.path or parsed.params or parsed.query or parsed.fragment:
+        return False
+
+    origin_default_port = 443 if parsed.scheme == "https" else 80
+    request_scheme = request_scheme.split(",", 1)[0].strip().lower()
+    request_default_port = 443 if request_scheme in ("https", "wss") else 80
+    origin_host = _host_port_from_netloc(parsed.netloc, origin_default_port)
+    request_host = _host_port_from_netloc(host or "", request_default_port)
+    if origin_host is None or request_host is None:
+        return False
+    if not _is_loopback_host(request_host[0]):
+        return False
+    return origin_host == request_host
+
+
+def _reject_untrusted_origin(request: web.Request) -> web.Response | None:
+    """Reject browser-origin WebSocket upgrades that would expose local CDP."""
+    origin = request.headers.get("Origin")
+    host = request.headers.get("Host")
+    scheme = request.headers.get("X-Forwarded-Proto", getattr(request, "scheme", "http"))
+    if _origin_is_allowed(origin, host, request_scheme=scheme):
+        return None
+    logger.warning("Rejected CDP WebSocket from untrusted Origin %r for Host %r", origin, host)
+    return web.Response(status=403, text="Forbidden: untrusted WebSocket origin\n")
 
 
 # ---------------------------------------------------------------------------
@@ -527,8 +613,12 @@ async def proxy_cdp_websocket(
         logger.error("%s error: %s", label, exc)
 
 
-async def handle_ws_default(request: web.Request) -> web.WebSocketResponse:
+async def handle_ws_default(request: web.Request) -> web.StreamResponse:
     """WebSocket proxy for default (no-seed) Chrome: /devtools/{type}/{guid}"""
+    rejected = _reject_untrusted_origin(request)
+    if rejected is not None:
+        return rejected
+
     pool: ChromePool = request.app["pool"]
     path = request.match_info.get("path", "")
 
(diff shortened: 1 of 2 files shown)


Cloak-HQ:
Thanks for the clean implementation. Origin guard works as expected — tested manually in Docker (6 scenarios: no origin, loopback match, attacker origin, public host match, DevTools origin, null origin) plus the full test suite (82 passed).

Good call on the strict host:port matching and rejecting before `get_or_launch()` so untrusted origins never spin up a Chrome process.

Merged via squash.

Issue #243, opened 2026-05-15
google.com not loading when using proxy


vladiulianbogdan:
Hello,

Firstly, thank you very much for this project!

I am testing Cloak Browser now for my use case and I've executed this script:

```
from cloakbrowser import launch
import time

browser = launch(proxy="<my_proxy_url>", headless=False, geoip=True, args=["--disable-gpu"])
page = browser.new_page()
page.goto("https://google.com", timeout=10000)  # no more blocks
time.sleep(3000)
```

The page does not load at all, it is stuck in loading. If I run a normal Chromium with the same proxy, it works. I've also tried to change the proxy provider and it still does not work. 

Am I doing something wrong or it's an issue somewhere. I am happy to debug this, let me know, what info do you need. I am running this on MacOS right now, but I've tested it also on a Docker environment with Linux and it was the same issue.

{'version': '145.0.7632.109.2', 'bundled_version': '146.0.7680.177.3', 'platform': 'darwin-arm64', 'download_url': 'https://cloakbrowser.dev/chromium-v145.0.7632.109.2/cloakbrowser-darwin-arm64.tar.gz'}
cloakbrowser python version: 0.3.28

vladiulianbogdan:
I've managed to make it work using this in args:

```
"--disable-features=ClientHintsMetaEquivalentHTTPEquivAcceptCH,CriticalClientHint,AcceptCHFrame"
```

It seems to be something related to the ACCEPT_CH frame and Chromium's "Critical Client Hints" mechanism. Is there a bug in CloakBrowser related to this?

Cloak-HQ:
Thanks for the report and nice find on the root cause.

This is Chromium's ACCEPT_CH HTTP/2 frame mechanism. Google (and other sites) send an ACCEPT_CH frame requesting Client Hints, and Chromium's `CriticalClientHint` feature restarts the connection to include them. With certain proxy setups, this restart loop hangs indefinitely. #117 was likely the same issue — `--disable-http2` worked there because it removed ACCEPT_CH frames as a side effect.

The cleanest fix is:

```python
browser = launch(
    proxy="your_proxy_url",
    headless=False,
    geoip=True,
    args=["--disable-features=AcceptCHFrame,CriticalClientHint"]
)
```

This disables the ACCEPT_CH frame handling that causes the hang, while keeping HTTP/2 working normally. Your `--disable-features` list also works but includes more flags than needed.

Let us know if that resolves it.

[...]

PR #247, opened 2026-05-15
fix(browser): disable critical client hints to prevent proxy hangs #243

siddhant-bayas:
# fix: add flags to prevent browser hang on proxy or geoip initialization (#243 )

## summary:
- adds --disable-features for Critical Client Hints and AcceptCH frames.
- resolves an issue where page.goto() hangs indefinitely on MacOS and Linux.

## justification:
chromium's internal retry logic for critical hints causes a deadlock with low-level fingerprinting patches during the initial proxy handshake. Disabling these features bypasses the problematic code path without impacting general automation functionality.

### DIFF
--- a/cloakbrowser/browser.py
+++ b/cloakbrowser/browser.py
@@ -958,17 +958,16 @@ def build_args(
     locale: str | None = None,
     headless: bool = True,
 ) -> list[str]:
-    """Combine stealth args with user-provided args and locale flags.
-
-    Deduplicates by flag key (everything before '=').
-    Priority: stealth defaults < user args < dedicated params (timezone/locale).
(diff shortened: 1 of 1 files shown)


Cloak-HQ:
Thanks for the PR, @siddhant-bayas.

We've fixed this from the other direction — the upcoming release routes authenticated HTTP proxy credentials through Chrome's `--proxy-server` flag instead of Playwright's CDP auth interceptor. That eliminates the 407 round-trip that was causing the hang, so the `--disable-features` workaround is no longer needed.

Closing this one, but appreciate the effort. The issue itself (#243) will be closed with the next release.

Issue #217, opened 2026-05-11
[Security][CVE request] Unauthenticated cloakserve CDP endpoint allows path traversal leading to arbitrary directory deletion


0xlally:
## Summary

`cloakserve` exposes an unauthenticated CDP multiplexer and uses the attacker-controlled `fingerprint` query parameter directly as the Chrome process/profile key. That value flows into the profile path and is later removed with `shutil.rmtree()`.

This allows a remote unauthenticated attacker who can reach `cloakserve` to perform path traversal outside the configured `data_dir`, resulting in deletion of attacker-selected directories reachable by the `cloakserve` process user. The same unauthenticated surface also returns CDP `webSocketDebuggerUrl` metadata.

I am filing this as a CVE candidate / security issue.

## Affected Version

Confirmed on `cloakbrowser 0.3.27`.

This may have been introduced in `0.3.20`, when `cloakserve` was rewritten as a CDP multiplexer with per-connection fingerprint seeds, but I have not bisected the exact introducing commit.

## Impact

If `cloakserve` is reachable over the network, an attacker can:

- Use `fingerprint=../...` or absolute path forms to make `user_data_dir` resolve outside `data_dir`.
- Trigger cleanup paths that call `shutil.rmtree(proc.user_data_dir, True)`.
- Delete directories accessible to the service user.
- Access unauthenticated CDP metadata and, in exposed deployments, connect to the CDP WebSocket.
- Create many unique `fingerprint` values to spawn Chrome processes, causing resource exhaustion.

Suggested severity: High/Critical depending on deployment exposure.

Suggested CVSS when `cloakserve` is network-accessible:

`CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:H` = 9.8

## Vulnerable Code Flow

Relevant code is in `bin/cloakserve`:

1. `parse_connection_params()` copies `fingerprint` into `result["seed"]`.
2. `ChromePool.get_or_launch()` assigns it to `seed_key`.
3. `user_data_dir = os.path.join(self._data_dir, seed_key)`.
4. The Chrome args include `--user-data-dir={user_data_dir}`.
5. Cleanup calls `shutil.rmtree(proc.user_data_dir, True)`.

There is no validation that the resolved `user_data_dir` remains under `data_dir`.

## Route-Level Reproduction

[...]

PR #248, opened 2026-05-15
fix(cloakserve): opt-in --auth-token + explicit --host for the CDP surface (#217)

therahul-yo:
## Summary

Addresses the remaining gaps in #217 after the path-traversal fix (commit `babef04`).

The reporter flagged three classes of risk; one was fixed, two remained:

| Gap | Status before this PR | After |
|---|---|---|
| Path traversal via `fingerprint=..%2F...` → arbitrary `shutil.rmtree` | ✅ Fixed in `babef04` (`SAFE_SEED_RE` + `_safe_rmtree`) | — |
| Unauthenticated `/json/version`, `/json/list`, `/devtools/...`, `/fingerprint/.../devtools/...` | ❌ Wide open | ✅ Opt-in `--auth-token` (Bearer or `?token=`) |
| `cloakserve` auto-binds `0.0.0.0` in containers — Docker port-forward exposes the CDP surface by default | ❌ Implicit non-loopback exposure | ✅ Explicit `--host` flag + startup warning when bound non-loopback with no token |

## Why this shape

- **Backwards-compatible.** No token = no auth = current behaviour. Existing `connect_over_cdp(...)` users keep working unchanged.
- **Two ways to pass the token.** `Authorization: Bearer <secret>` (preferred — keeps the secret out of access logs) or `?token=<secret>` (compat for clients that can't set headers, e.g. Playwright `connect_over_cdp` URL strings).
- **Env-var fallback** (`CLOAKSERVE_AUTH_TOKEN`) so the secret doesn't have to land in process-list output.
- **Constant-time compare** via `hmac.compare_digest` — the endpoint can't be used as a length/prefix oracle.
- **`/` (status) stays open** so container health checks keep working without leaking the token to every probe.
- **Loud startup warning** when the resolved bind is non-loopback AND no auth token is set. This is the realistic remediation path for existing deployments — they keep working but get a clear nudge toward the flag.

## What's NOT in this PR

- **Per-IP / per-seed rate limiting** to bound the Chrome-process spawn DoS — left for a follow-up. Wanted to keep this PR focused on auth + bind.
- **Default-secure bind change** — not flipping the `0.0.0.0`-in-container default to keep behaviour stable for current users. The warning gives operators time to opt into `--auth-token`.

## Test plan

12 new tests covering CLI parsing and middleware behaviour:

| Test class | Covers |
|---|---|
| `TestHostAndAuthCli` (6) | `--host`, `--auth-token`, `CLOAKSERVE_AUTH_TOKEN`, CLI-beats-env precedence |
| `TestAuthMiddleware` (8) | no-token = open; `/` always open; missing/wrong/correct bearer; query-param token; `/devtools/*` and `/fingerprint/*/devtools/*` both protected |

```
$ python -m pytest tests/test_cloakserve.py -v
======================== 66 passed, 8 warnings in 0.12s ========================
```

## Files

- `bin/cloakserve` — `+45 / -3` (middleware, two CLI flags, env fallback, startup warning)
- `tests/test_cloakserve.py` — `+185 / 0` (regression coverage)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/bin/cloakserve
+++ b/bin/cloakserve
@@ -325,8 +325,16 @@ class ChromePool:
 # Query param parsing
 # ---------------------------------------------------------------------------
 
-# Params that need special handling (not simple --fingerprint-{name}= mapping)
-SPECIAL_PARAMS = {"fingerprint", "proxy", "geoip", "locale", "timezone"}
+# Params that need special handling (not simple --fingerprint-{name}= mapping).
+# ``token`` is reserved for the ``?token=<secret>`` auth flow (see
+# ``auth_middleware``).  It MUST be excluded from the generic branch below —
+# otherwise the value would fall through and get forwarded into Chrome's
+# args as ``--fingerprint-token=<secret>``, leaking the auth secret to
+# child process argv (visible in ``ps``) and emitting an unsupported flag.
+# Surfaced in review of #217 fix.
+SPECIAL_PARAMS = {
+    "fingerprint", "proxy", "geoip", "locale", "timezone", "token",
+}
 
 
 def parse_connection_params(query_string: str) -> dict:
@@ -418,6 +426,24 @@ async def auth_middleware(
     return await handler(request)
 
 
+def _append_token_query(url: str, token: str | None) -> str:
+    """Append ``?token=<token>`` to a rewritten WebSocket URL when auth is on.
+
+    The middleware accepts auth via either ``Authorization: Bearer`` or
+    ``?token=`` (compat for ``connect_over_cdp(URL_STRING)`` clients that
+    cannot set request headers).  After ``/json/version`` or ``/json/list``
+    rewrites the WebSocket URL, the client follows that URL to open a WS
+    handshake — without the token propagated onto the WS URL, the auth
+    middleware rejects the handshake with 401 and breaks the URL-string
+    use case. Surfaced in review of #217 fix.
+    """
+    if not token:
+        return url
+    sep = "&" if "?" in url else "?"
+    from urllib.parse import quote
+    return f"{url}{sep}token={quote(token, safe='')}"
+
+
 def _ws_scheme(request: web.Request) -> str:
     """Return 'wss' if client connected via HTTPS (e.g. TLS-terminating proxy), else 'ws'."""
     proto = request.headers.get("X-Forwarded-Proto", request.scheme)
@@ -484,7 +510,10 @@ async def handle_json_version(request: web.Request) -> web.Response:
     guid = orig_ws.rsplit("/", 1)[-1] if "/devtools/" in orig_ws else ""
 
     scheme = _ws_scheme(request)
-    data["webSocketDebuggerUrl"] = f"{scheme}://{host}/{ws_path}/{guid}"
+    auth_token = request.app.get("auth_token")
+    data["webSocketDebuggerUrl"] = _append_token_query(
+        f"{scheme}://{host}/{ws_path}/{guid}", auth_token,
+    )
     return web.json_response(data)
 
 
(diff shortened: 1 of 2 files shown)


journeytosilius (changes requested):
First-pass review. I found two blockers in the query-token auth path:

1. `bin/cloakserve:332-359` does not exclude `token` from `parse_connection_params()`. A request like `/json/version?token=s3cret` authenticates in the middleware, then the handler parses the same query string and treats `token` as a generic fingerprint parameter, forwarding `--fingerprint-token=s3cret` through `extra_args` into the Chrome launch path. That leaks the auth secret into child process args and may also pass an unsupported flag. Please strip `token` from the launch-param parser, e.g. by adding it to `SPECIAL_PARAMS`, and cover this with a regression test.

2. `bin/cloakserve:487` and `bin/cloakserve:524-528` rewrite `webSocketDebuggerUrl` without preserving the query token. With `auth_token` configured, a client using the advertised `?token=` compatibility can authenticate `/json/version` or `/json/list`, but the returned `ws://.../devtools/...` URL has no token, so the new middleware rejects the subsequent WebSocket handshake with 401. That breaks the stated Playwright `connect_over_cdp` URL-string use case. Please propagate the token into the rewritten WebSocket URLs, or otherwise make the discovered WS URL authenticate, and add an end-to-end test for that flow.

Local validation: attempted `python3 -m pytest tests/test_cloakserve.py -q`, but this environment is missing `pytest` (and `websockets`), so I reviewed against the exported PR tree and diff instead.

therahul-yo:
Thanks for the careful review @journeytosilius — both blockers are real and now fixed in 91b1230.

**Blocker 1 — `?token=` leaking into Chrome args.**
Added `token` to `SPECIAL_PARAMS` so `parse_connection_params` silently drops it instead of falling through to the generic `--fingerprint-{key}={val}` branch.
- New test `test_token_query_param_is_not_forwarded_to_chrome` exercises the exact `fingerprint=1&token=super-secret-token-value` shape from your example and asserts no `token` substring appears anywhere in `extra_args`.
- New test `test_token_query_param_alone_is_silently_consumed` covers the bare `?token=abc` case.

**Blocker 2 — rewritten `webSocketDebuggerUrl` lost the token.**
New `_append_token_query(url, token)` helper appends `?token=<urlquoted-token>` (or `&token=...` if a query already exists) when `app["auth_token"]` is configured. Wired into both rewrite sites — `handle_json_version` at L495 and `handle_json_list` at L528 — so the Playwright `connect_over_cdp(URL_STRING)` compat flow round-trips end-to-end.
- `TestAppendTokenQuery` (5 unit cases): no-token no-op, no-query append, with-query append, URL encoding, fingerprint path.
- `TestHandleJsonVersionTokenPropagation` (2 end-to-end cases) spins up a stub Chrome `/json/version` upstream and asserts the rewritten WS URL carries `?token=rev-secret` when auth is configured, and is clean when it isn't.

Full suite: **75 passed** (66 prior + 9 new). To reproduce locally:
```bash
pip install -e . pytest pytest-asyncio aiohttp websockets
python -m pytest tests/test_cloakserve.py -v
```

Cloak-HQ:
Thanks for the thorough PR! There's a merge conflict with `bin/cloakserve` from #240 (WebSocket origin guards) that landed on main. Could you rebase on main and resolve it? Let us know if you need a hand.

rafaelfiguereod-stack:
Nice — opt-in `--auth-token` is the right shape here.

One complementary gap that's orthogonal to auth and might be worth folding in (or as a follow-up): the `/` status handler returns each session's `proxy` string verbatim, and the "seed already running" `logger.warning(...)` logs it too — both include inline `user:pass` credentials when a session was started with an authenticated proxy (e.g. `?proxy=http://user:pass@host:8080`).

`--auth-token` gates *who* can reach `/`, but an authorized operator — and the log file / any log aggregation — still sees plaintext proxy creds, which for paid residential/ISP proxies is a real secret. So it's defense-in-depth alongside, not instead of, the token.

A tiny `_redact_proxy()` that masks the userinfo (→ `http://***@host:8080`), applied in `handle_root` and that warning, closes it with no behavior change. Happy to send a small PR (against this branch or standalone) if that's useful.

PR #257, opened 2026-05-17
refactor: export buildLaunchOptions and humanizeBrowser from JS package

### LINKED ISSUE #241: [JS] Export function to only build options
It would be useful if the npm package exported functions to build options and humanize browser.

```typescript
export async function buildLaunchOptions(options: LaunchOptions = {}): Promise<LaunchOptions> {
  const binaryPath = process.env.CLOAKBROWSER_BINARY_PATH || (await ensureBinary());
  const { exitIp, ...resolved } = await maybeResolveGeoip(options);
  const { proxyOption, proxyArgs } = resolveProxyConfig(options.proxy);
  let resolvedArgs = await resolveWebrtcArgs(options);
  if (exitIp && !(resolvedArgs ?? []).some(a => a.startsWith("--fingerprint-webrtc-ip"))) {
    resolvedArgs = [...(resolvedArgs ?? []), `--fingerprint-webrtc-ip=${exitIp}`];
  }
  const args = buildArgs({ ...options, ...resolved, args: [...(resolvedArgs ?? []), ...proxyArgs] });

  return {
    executablePath: binaryPath,
    headless: options.headless ?? true,
    args,
    ignoreDefaultArgs: IGNORE_DEFAULT_ARGS,
    ...(proxyOption ? { proxy: proxyOption } : {}),
    ...options.launchOptions,
  };
}

export async function humanizeBrowser(browser: Browser, options: LaunchOptions = {}): Promise<Void> {
  if (options.humanize) {
    const { patchBrowser } = await import('./human/index.js');
    const { resolveConfig } = await import('./human/config.js');
    const cfg = resolveConfig(
      options.humanPreset ?? 'default',
      options.humanConfig,
    );
    patchBrowser(browser, cfg);
  }
}

export async function launch(options: LaunchOptions = {}): Promise<Browser> {
  const { chromium } = await import("playwright-core");
  const browser = await chromium.launch(await buildLaunchOptions(options));
  await humanizeBrowser(browser, options);
  return browser;
}
```

---

The npm package exports functions that override Playwright. It’s difficult to integrate it with other tools that also modify Playwright. By exporting the functions used to build options and humanize the browser, we have more flexibility. For example, we can use a modified version of Playwright.

```javascript
import { buildLaunchOptions, humanizeBrowser } from "cloakbrowser";
import { chromium } from "./my-awesome-playwright-custom.js";

const browser = await chromium.launch(await buildLaunchOptions());
```

The feature seems similar to the issue https://github.com/CloakHQ/CloakBrowser/issues/153 for Python.

The [Camoufox JavaScript port](https://github.com/apify/camoufox-js#usage) exports the `launchOptions()` function.

---

I'd like to integrate CloakBrowser into [Playwright-ghost](https://github.com/regseb/playwright-ghost), but it's difficult without these new functions.

mvanhorn:
## Summary

Exposes `buildLaunchOptions` and `humanizeBrowser` as named exports from the JS `cloakbrowser` package so callers can compose the launch pipeline against a user-supplied Playwright variant. `launch()` keeps its existing behavior and now delegates to the two helpers.

## Why this matters

In #241 @regseb (author of Playwright-ghost) asked for lower-level builder helpers because today `launch()` always calls `await import("playwright-core")` internally, so integrating CloakBrowser with a modified Playwright build or another tool that also patches Playwright requires inline-copying the body of `launch()` per release. The proposal mirrors the Python equivalent ask in #153 and the API surface that Camoufox's apify port already provides.

The PR uses the exact API shape @regseb sketched in the issue: a `buildLaunchOptions(options)` builder that returns a `LaunchOptions` object you can pass to your own `chromium.launch()`, plus a separate `humanizeBrowser(browser, options)` step for the post-launch side effect.

## Changes

- `buildLaunchOptions(options)` returns the assembled `LaunchOptions` (`executablePath`, `headless`, `args`, `ignoreDefaultArgs`, optional `proxy` + `launchOptions` overrides). No proxy key is set when one is not provided.
- `humanizeBrowser(browser, options)` performs the existing `patchBrowser` side effect, and is a no-op when `options.humanize` is falsy (does not import `./human/index.js`).
- `launch()` is rewritten as `chromium.launch(await buildLaunchOptions(options))` followed by `await humanizeBrowser(browser, options)`, with both helpers re-exported from `js/src/index.ts`.

## Testing

- Added `js/tests/build-launch-options.test.ts` (8 unit tests, all passing). Uses `stealthArgs: false` and a fixed `--fingerprint=12345` so the args snapshot is deterministic.
- `npm run typecheck` clean.
- Full vitest suite (340 tests) passes; no regression in the existing `launch()` flow.

Closes #241

### DIFF
--- a/js/src/index.ts
+++ b/js/src/index.ts
@@ -16,7 +16,7 @@
  */
 
 // Launch functions (Playwright API)
-export { launch, launchContext, launchPersistentContext } from "./playwright.js";
+export { launch, launchContext, launchPersistentContext, buildLaunchOptions, humanizeBrowser } from "./playwright.js";
 
 // Binary management
 export { ensureBinary, clearCache, binaryInfo, checkForUpdate } from "./download.js";
--- a/js/src/playwright.ts
+++ b/js/src/playwright.ts
@@ -45,6 +45,72 @@ function filterStealthCtxOptions(ctx?: BrowserContextOptions): Partial<BrowserCo
 }
 
 /**
+ * Build the LaunchOptions object that {@link launch} would forward to
+ * `chromium.launch()`. Use this when you want to drive your own
+ * `playwright-core` (or a modified Playwright build) and only need the
+ * cloakbrowser launch-arg pipeline — binary path, geoip resolution, proxy
+ * resolution, WebRTC arg resolution, and stealth arg merging.
+ *
+ * @example
+ * ```ts
+ * import { chromium } from 'playwright-core';
+ * import { buildLaunchOptions } from 'cloakbrowser';
+ *
+ * const opts = await buildLaunchOptions({ headless: false });
+ * const browser = await chromium.launch(opts);
+ * ```
+ */
+export async function buildLaunchOptions(
+  options: LaunchOptions = {}
+): Promise<Record<string, unknown>> {
+  const binaryPath = process.env.CLOAKBROWSER_BINARY_PATH || (await ensureBinary());
+  const { exitIp, ...resolved } = await maybeResolveGeoip(options);
+  const { proxyOption, proxyArgs } = resolveProxyConfig(options.proxy);
+  let resolvedArgs = await resolveWebrtcArgs(options);
+  if (exitIp && !(resolvedArgs ?? []).some(a => a.startsWith("--fingerprint-webrtc-ip"))) {
+    resolvedArgs = [...(resolvedArgs ?? []), `--fingerprint-webrtc-ip=${exitIp}`];
+  }
+  const args = buildArgs({ ...options, ...resolved, args: [...(resolvedArgs ?? []), ...proxyArgs] });
+
+  return {
+    executablePath: binaryPath,
+    headless: options.headless ?? true,
+    args,
+    ignoreDefaultArgs: IGNORE_DEFAULT_ARGS,
+    ...(proxyOption ? { proxy: proxyOption } : {}),
+    ...options.launchOptions,
+  };
+}
+
+/**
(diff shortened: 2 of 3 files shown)


regseb:
You should do the same for the `launchPersistentContext()` function.

And for `launchContext`:

```typescript
export async function launchContext(
  options: LaunchContextOptions = {}
): Promise<BrowserContext> {
  const browser = await launch(options);

  let context: BrowserContext;
  try {
    context = await browser.newContext(await buildContextOptions(options));
  } catch (err) {
    await browser.close();
    throw err;
  }

  // Patch close() to also close the browser
  const origClose = context.close.bind(context);
  context.close = async () => {
    await origClose();
    await browser.close();
  };

  await humnizeContext(context, options);
  return context;
}
```

I think there's a way to factor out the code by renaming `buildLaunchOptions` to `buildBrowserOptions`.

- `launch`: `buildBrowserOptions` & `humanizeBrowser`
- `launchContext`: `launch`, `buildContextOptions` & `humanizeContext`
- `launchPersistentContext`: `buildBrowserOptions`, `buildContextOptions` & `humanizeContext`

Cloak-HQ:
Closing — this was already merged via #244. Thanks for the contribution though, @mvanhorn!

mvanhorn:
@regseb - opened #265 with the same compose-style refactor for `launchContext` and `launchPersistentContext`. Exports `buildContextOptions`, `humanizeContext`, `buildPersistentContextOptions` and refactors the entry functions to use them. Thanks for the suggestion!

Issue #253, opened 2026-05-17
Add option to install extension when the profile is created, like gologin


ojtvpn:

Cloak-HQ:
This is supported via the `extension_paths` parameter, shipping now in v0.3.29:

```python
context = cloakbrowser.launch_persistent_context(
    user_data_dir="./my-profile",
    extension_paths=["./my-extension", "./another-extension"],
    headless=False,
)
```

Pass a list of paths to unpacked extension directories. Added in #210.

PR #258, opened 2026-05-17
Feature: Profile Extension Pre-installation & Injection

siddhant-bayas:
###  Core Enhancements
- **[Feature] Automated Profile Extension Injection (#253 ):** Added GoLogin-like parity allowing developers to attach Chrome Web Store extensions directly to individual profiles at creation. Features on-the-fly `.crx` downloading, sandbox caching, and automatic persistent context launching arguments setup. Contributed comprehensive documentation and diagnostic guidelines in the project README.

 ### Technical Contribution:
- Automated CRX Pipeline: Created a Python network utility to seamlessly download extension packages (.crx/.zip) straight from the official Chrome Web Store using standard extension IDs.

- Filesystem Sandboxing: Added automated decompression handling that isolates extracted extension code within an isolated extensions/{extension_id} subdirectory inside each specific profile's user data path.

- Runtime Flag Injection: Patched the profile browser initiation process to look for saved extensions, automatically bundle their absolute paths, and dynamically inject them via --load-extension and --disable-extensions-except arguments inside launch_persistent_context().

- Anti-Detection Lifecycle Guardrail: Forced a runtime switch to headless=False whenever extensions are active to protect extension rendering pipelines and maintain humanization protocols.

### Documentation & DX (Developer Experience) Improvements

- Comprehensive Integration Guide: Authored and integrated an extensive section into the project README.md explaining how the extension manager handles execution.

- User Optimization Mapping: Documented expected performance deltas.

- Troubleshooting Guide: Provided clear instructions on scraping extension IDs directly from Chrome Web Store URLs alongside diagnostic steps for debugging failing or corrupted .crx unzipping operations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -377,6 +377,49 @@ Supports all the same options as `launch_context()`: `proxy`, `user_agent`, `vie
 
 Async version: `launch_persistent_context_async()`.
 
+#### Chrome Extensions
+
+CloakBrowser supports automatic installation and injection of Chrome extensions when launching a persistent profile — just like anti-detect browsers (GoLogin, etc.).
+
+Extensions are downloaded from the Chrome Web Store, extracted, and stored in the profile. They persist across sessions and are automatically loaded on each launch.
+
+```python
+from cloakbrowser import launch_persistent_context
+
+# Install MetaMask and uBlock Origin
+ctx = launch_persistent_context(
+    "./my-profile",
+    extensions=[
+        "nkbihfbeogaeaoehlefnkodbefgpgknn",  # MetaMask
+        "cjpalhdlnbpafiamejdnhcphjbkeiagm",  # uBlock Origin
+    ],
+    # headless=True  (automatically set to False; extensions don't run headless)
+)
+
+page = ctx.new_page()
+page.goto("https://app.uniswap.org")
+
+# MetaMask is now available
+has_ethereum = page.evaluate("typeof window.ethereum !== 'undefined'")
+print(f"MetaMask detected: {has_ethereum}")
+
+ctx.close()
+```
+
+**How it works:**
+- Extension IDs are standard Chrome Web Store IDs (e.g., find them at `https://chrome.google.com/webstore/`)
+- On first launch, extensions are downloaded and extracted to `{profile_dir}/extensions/{extension_id}/`
+- On subsequent launches with the same profile, cached extensions are used (no re-download)
+- Headless mode is automatically disabled (`headless=False`) — extensions only run in headed mode
+- Extensions are loaded with `--load-extension` and `--disable-extensions-except` flags
+
+**Error handling:**
+- Invalid extension IDs are logged and skipped (browser still launches)
+- Network timeouts or corrupted archives are caught and logged
+- Extensions are cached — if download fails, the cached version is used next time
+
+**Storage quota:** Extensions increase the apparent storage quota (quota check is one factor in browser fingerprinting). If you're using FingerprintJS or similar detection, this may need adjustment — see the storage quota section above.
+
 **Storage quota and detection tradeoff:** By default, the binary normalizes storage quota to pass FingerprintJS, which blocks persistent contexts that report non-incognito quota values. This means detection services that penalize incognito mode (like BrowserScan's `notPrivate` check, -10 points) will still flag it. If your target site penalizes incognito but doesn't use FingerprintJS, set a higher quota to appear as a regular profile:
 
 ```python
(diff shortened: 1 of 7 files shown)


Cloak-HQ:
Thanks for putting this together, and for the thorough write-up.

Heads up on overlap: we shipped local extension loading in #210 (merged May 15) — `extension_paths=[...]` is now on all launch functions (`launch`, `launch_context`, `launch_persistent_context`, plus async variants), and `build_args` handles the `--load-extension` / `--disable-extensions-except` injection. So the flag-injection half of this PR is already on main, and the branch will conflict.

The genuinely new piece here is the auto-download from the Chrome Web Store (`extensions.py`) — a nice addition on top of `extension_paths`. If you'd like to pursue it, a few things would need to change:

1. **Build on `extension_paths`, don't re-inject flags.** Resolve CWS IDs → local paths in `extensions.py`, then pass those through the existing `extension_paths` param. That keeps it working across all launch functions, not just persistent context.
2. **Drop the `--disable-features` / client-hints change.** It's unrelated to extensions (looks tied to #243) and touches the shared arg builder — should be its own PR.
3. **Fix the CRX3 extraction.** The fixed `[16, 12, 4]` offset skips don't reliably locate the ZIP in CRX3 — the header length is variable and should be read from the header (bytes 8–12), not guessed.
4. **Reconsider the silent `headless=False`.** Forcing headed breaks headless servers; a warning or explicit opt-in is safer.
5. **Trim the README.** The current addition is ~270 lines; a feature like this needs a short usage block (snippet + a line on caching), not a full tutorial with performance tables and troubleshooting.

Let us know if you'd like to rework it along those lines — happy to review again.

siddhant-bayas:
> Thanks for putting this together, and for the thorough write-up.
> 
> Heads up on overlap: we shipped local extension loading in #210 (merged May 15) — `extension_paths=[...]` is now on all launch functions (`launch`, `launch_context`, `launch_persistent_context`, plus async variants), and `build_args` handles the `--load-extension` / `--disable-extensions-except` injection. So the flag-injection half of this PR is already on main, and the branch will conflict.
> 
> The genuinely new piece here is the auto-download from the Chrome Web Store (`extensions.py`) — a nice addition on top of `extension_paths`. If you'd like to pursue it, a few things would need to change:
> 
>     1. **Build on `extension_paths`, don't re-inject flags.** Resolve CWS IDs → local paths in `extensions.py`, then pass those through the existing `extension_paths` param. That keeps it working across all launch functions, not just persistent context.
> 
>     2. **Drop the `--disable-features` / client-hints change.** It's unrelated to extensions (looks tied to [google.com not loading when using proxy #243](https://github.com/CloakHQ/CloakBrowser/issues/243)) and touches the shared arg builder — should be its own PR.
> 
>     3. **Fix the CRX3 extraction.** The fixed `[16, 12, 4]` offset skips don't reliably locate the ZIP in CRX3 — the header length is variable and should be read from the header (bytes 8–12), not guessed.
> 
>     4. **Reconsider the silent `headless=False`.** Forcing headed breaks headless servers; a warning or explicit opt-in is safer.
> 
>     5. **Trim the README.** The current addition is ~270 lines; a feature like this needs a short usage block (snippet + a line on caching), not a full tutorial with performance tables and troubleshooting.
> 
> 
> Let us know if you'd like to rework it along those lines — happy to review again.

Thanks for the detailed review and for pointing out the overlap with #210.

That makes sense — I missed that extension_paths had already been generalized across all launch APIs, so re-injecting the extension flags in this PR is redundant and would definitely create conflicts with the current arg builder flow.

I’ll rework the implementation around extension_paths instead:

* keep extensions.py focused on resolving/downloading CWS extensions into local directories
* feed the resolved paths through the existing extension_paths parameter
* remove the custom flag injection entirely

I’ll also split out the unrelated --disable-features / client hints change. You're right that it doesn’t belong in an extensions PR, especially since it modifies shared launch args and is tied to separate proxy behavior.

Good catch on the CRX3 parsing too. The fixed offset approach was naïve in hindsight; I’ll switch to reading the CRX3 header length properly from the metadata section instead of assuming static skips. Humans really did invent at least six archive formats just to store ZIP files inside other ZIP files. Remarkable species.

I’ll remove the implicit headless=False behavior as well and replace it with either:

* documentation/warnings about extensions in headless mode, or
* an explicit opt-in if needed

And agreed on the README. I overcooked it. I’ll trim it down to a concise usage example + caching behavior and move implementation details out of the main docs.

I’ll push a cleaned-up revision aligned with the existing architecture and reopen for review once it’s ready.

PR #262, opened 2026-05-18
feat(js): export buildContextOptions helper

honor2030:
## Summary
- Export `buildContextOptions()` from the JS package entrypoint for existing Playwright Browser integrations
- Reuse the helper inside `launchContext()` and `launchPersistentContext()` so the composable path matches wrapper defaults
- Preserve stealth-safe context option filtering for `locale` and `timezoneId`, plus default viewport behavior

Refs #241

## Test Plan
- RED: `PATH=/opt/homebrew/opt/node@20/bin:/opt/homebrew/bin:$PATH npm test -- --run tests/launch.test.ts --reporter=verbose` failed with `buildContextOptions is not a function`
- GREEN: `PATH=/opt/homebrew/opt/node@20/bin:/opt/homebrew/bin:$PATH npm test -- --run tests/launch.test.ts --reporter=verbose` → `22 passed | 1 skipped`
- `PATH=/opt/homebrew/opt/node@20/bin:/opt/homebrew/bin:$PATH npm run typecheck`
- `PATH=/opt/homebrew/opt/node@20/bin:/opt/homebrew/bin:$PATH npm test -- --run --reporter=dot` → `10 passed (10), 326 passed | 11 skipped`
- `PATH=/opt/homebrew/opt/node@20/bin:/opt/homebrew/bin:$PATH npm run build`
- `git diff --check`

PR #265, opened 2026-05-18
feat(js): export composable context helpers

mvanhorn:
Mirrors the composable launch helpers from #244 for the context layer. Surface change:

- `buildContextOptions(options)` returns the `BrowserContextOptions` object that `launchContext()` previously assembled inline.
- `humanizeContext(context, options)` applies the human-like layer to an existing context (counterpart to `humanizeBrowser`).
- `buildPersistentContextOptions(options)` returns the merged options object passed to `chromium.launchPersistentContext()`.

`launchContext` and `launchPersistentContext` are refactored to use the new helpers. Observable behavior is unchanged.

Follow-up to @regseb's suggestion on #257. Three new tests in `js/tests/launch.test.ts` cover the new helpers; `npm run typecheck`, `npm run build`, and `npm test` (327 passing, 11 skipped) all clean locally.

### DIFF
--- a/js/src/index.ts
+++ b/js/src/index.ts
@@ -16,7 +16,16 @@
  */
 
 // Launch functions (Playwright API)
-export { launch, launchContext, launchPersistentContext, buildLaunchOptions, humanizeBrowser } from "./playwright.js";
+export {
+  launch,
+  launchContext,
+  launchPersistentContext,
+  buildLaunchOptions,
+  humanizeBrowser,
+  buildContextOptions,
+  humanizeContext,
+  buildPersistentContextOptions,
+} from "./playwright.js";
 
 // Binary management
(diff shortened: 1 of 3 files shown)


Cloak-HQ:
Thanks for putting this together, @mvanhorn. We went ahead and merged #262 which covers the `buildContextOptions` helper with the same extraction.

One thing we spotted in this PR: `buildPersistentContextOptions` doesn't call `resolveTimezone` internally, so standalone callers passing `timezoneId` instead of `timezone` won't get the `--fingerprint-timezone` flag in the args. If you'd like to open a follow-up for `buildPersistentContextOptions` and `humanizeContext` separately, adding `resolveTimezone` at the top of that function would fix it.

Closing in favor of #262. Appreciate the contribution!

PR #291, opened 2026-05-21
Feature/custom search

sashahavrilyuk:
Added custom search engine for Chromium. Implemented by modifiyng the Chromium's default/preferences. Google, Bind or DuckDuckGo


Cloak-HQ:
Thanks for the PR! The implementation works — I verified it sets the default engine in Chromium's settings and persists across relaunches.

Before merging, could you share what your use case is? Since most wrapper users navigate programmatically (`page.goto()`), the omnibox default doesn't typically come into play. Curious what scenario prompted this — are you using it with the Manager (VNC), reusing profiles manually, or something else?

Also a couple of things to address:

1. `js/src/types.ts` declares `searchEngine` but there's no JS implementation — this would confuse TypeScript users. Either add the JS side or remove the type for now.
2. `launch_context` / `launch_context_async` accept the param and silently warn. Cleaner to omit it from those signatures entirely since it can never work without a user data dir.

sashahavrilyuk:
> Thanks for the PR! The implementation works — I verified it sets the default engine in Chromium's settings and persists across relaunches.
> 
> Before merging, could you share what your use case is? Since most wrapper users navigate programmatically (`page.goto()`), the omnibox default doesn't typically come into play. Curious what scenario prompted this — are you using it with the Manager (VNC), reusing profiles manually, or something else?
> 
> Also a couple of things to address:
> 
> 1. `js/src/types.ts` declares `searchEngine` but there's no JS implementation — this would confuse TypeScript users. Either add the JS side or remove the type for now.
> 2. `launch_context` / `launch_context_async` accept the param and silently warn. Cleaner to omit it from those signatures entirely since it can never work without a user data dir.

Yes, I’m using it with the profile manager, i also PR the manager for it https://github.com/CloakHQ/CloakBrowser-Manager/pull/26 , basically I’d like to add more functionality from other antidetect browsers like start page, cookie import/export and so on.

PR #301, opened 2026-05-23
docs: add Simplified Chinese translations for README

Robin-zero:


Cloak-HQ:
Thanks for the effort on the translation. We're keeping the project documentation in English only for now to keep maintenance manageable. If demand grows we may revisit with an i18n approach, but for now we'll pass.

Closing this one.

PR #303, opened 2026-05-23
fix: fixed checkPointerEventsHandle check

eofreternal:
Previously, there was an issue where locations inside iFrames were not offset correctly. Also: if the element that was covering the other element couldn't be determined, the click would fail which is no good. Now, the click is attempted anyway because otherwise, the browser would crash.

Not letting the click through a mistake. I encountered an error when trying to automate Google login's by using this code:
```
await page.goto("https://www.google.com/", { waitUntil: "networkidle" });
await page.click('[aria-label="Sign in"]')
typeInGoogleLogin();
```

It crashed saying that there was an element covering the Sign In button. There was not an element covering the Sign In button but CloakBrowser crashed anyway claiming there was. The real issue is that it couldn't determine if there was an element covering it or not. That can happen if there really is no element covering the button. If you're looking for an element covering the other element and you can't find it but you treat that as an error, well. you see how it could be bad.

Anyway, here's my fix for that

### DIFF
--- a/js/src/human/actionability.ts
+++ b/js/src/human/actionability.ts
@@ -197,8 +197,11 @@ export async function ensureStable(
 // Pointer-events check (post-scroll, at actual click coordinates)
 // ---------------------------------------------------------------------------
 
-const POINTER_EVENTS_LOCATOR_JS = `(expected, coords) => {
-  const target = document.elementFromPoint(coords.x, coords.y);
+const POINTER_EVENTS_LOCATOR_JS = `(expected, data) => {
+  const rect = expected.getBoundingClientRect();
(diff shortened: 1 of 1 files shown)


Cloak-HQ:
Thanks for the fix @eofreternal, this is merged.

You nailed the diagnosis. We reproduced it (an ElementHandle inside an iframe getting wrongly reported as covered, because the check used page-space coordinates inside the frame's own context) and confirmed your offset correction resolves it.

On our side we also ported the same fix to the Python wrapper (sync and async) to keep the two in parity, and added fail-open regression tests for both languages.

Let us know if you run into anything else, and thanks again for contributing.

PR #323, opened 2026-05-26
Readme updates

eofreternal:
Now, the important information is presented up at the top. I also got rid of some of the confusing language and presented everything in a cleaner manner


eofreternal:
@Cloak-HQ Ready for review!

PR #326, opened 2026-05-28
Support CommonJS in the JS package

murrayee:
## Summary

This PR adds CommonJS support for the `js` package while keeping the existing ESM exports.

## Changes

- Build both ESM and CommonJS outputs with `tsup`
- Add `require` entries to `exports` for:
  - `cloakbrowser`
  - `cloakbrowser/puppeteer`
  - `cloakbrowser/human`
- Set `main` to the CommonJS entrypoint
- Keep `module` pointing to the ESM entrypoint

## Why

The package already exposed ESM entrypoints, but it did not provide official CommonJS entrypoints through `exports`.

This change allows usage like:

```js
const { launch } = require('cloakbrowser');

### DIFF
--- a/js/package-lock.json
+++ b/js/package-lock.json
@@ -20,6 +20,7 @@
         "playwright-core": "1.61",
         "puppeteer-core": "^25.0.4",
         "socks-proxy-agent": "^10.0.0",
+        "tsup": "^8.5.1",
         "typescript": "^6.0.3",
         "vitest": "^3.2.6"
       },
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by Cloak-HQ** [comment]
Thanks for the PR, CommonJS support is something we want.

I built and tested the branch locally and hit two blockers that need fixing before merge.

**1. The CJS output doesn't load on the Node range we support.**

With `bundle: false`, esbuild keeps the NodeNext `.js` import specifiers as-is, so each emitted `.cjs` ends up requiring its ESM sibling:

```
dist/index.cjs → require("./playwright.js")   // .js is ESM, the .cjs sibling is never used
```

Because the package is `"type": "module"`, those `.js` files are ES modules, so `require()` throws:

```
ERR_REQUIRE_ESM: require() of ES Module dist/playwright.js from dist/index.cjs not supported
```

This affects all three new `require` entrypoints (`.`, `./puppeteer`, `./human`).
It fails on Node 18.x and on Node 20/22 before `require(esm)` was unflagged, so it breaks for most of our `engines: >=18` range.

Setting `bundle: true` in `tsup.config.ts` fixes it, I tested it and `require('cloakbrowser')` loads cleanly with `require(esm)` disabled.

**2. `npm run build` currently exits non-zero.**

The tsup DTS pass fails:

```
src/args.ts: TS2591: Cannot find name 'path' / 'process'
TS5101: Option 'baseUrl' is deprecated (TypeScript 6.x)
```

`@types/node` is installed and `tsc --noEmit` passes, so the DTS builder just isn't picking up the node types.
The side effect is that no `.d.ts` files get emitted, which leaves the `exports.types` entries pointing at missing files and drops types for TS consumers.

Could you flip to `bundle: true` and get the DTS build resolving node types so `npm run build` is green?
Happy to help if you get stuck on the DTS config.

murrayee:
@Cloak-HQ Thanks for the detailed review. I switched tsup to bundle: true and fixed the DTS config to resolve Node types and ignore the TS 6.0 deprecation warning. npm run build and npm test are both green locally now.

PR #352, opened 2026-06-05
fix(cloakserve): add idle cleanup for seeded profiles

### LINKED ISSUE #346: Persistant profiles leads to OOM
Hey I found this project recently and have been really liking it. One issue I encountered is that after some time my `docker.io/cloakhq/cloakbrowser` container would stop working. I did some troubleshooting and apparently the issue is each unique seed will create a new Chromium profile but it doesn't despawn.
Each new profile adds 200MB+ of RAM that eventually leads to OOM. As a result of this I am now only using the single `__default__` profile.

I checked the docs but couldn't find any mentions or disclaimers of:

- Profiles never despawn on disconnect
- There's no idle timeout, TTL, or automatic cleanup mechanism
- Memory grows linearly with unique seeds
- A container restart is the only way to free up memory

Is this the intended behaviour or am I doing something wrong?

Kumario1:
## Summary
- add an optional cloakserve idle cleanup timer for per-seed Chrome processes after the last CDP WebSocket disconnects
- expose idle cleanup state in GET / and support both --idle-timeout=SECONDS and CLOAKSERVE_IDLE_TIMEOUT
- document the memory-bounding option and profile cleanup behavior

Fixes #346

## Tests
- python -m pytest tests/test_cloakserve.py -q
- python -m compileall -q bin/cloakserve tests/test_cloakserve.py
- git diff --check

### DIFF
--- a/README.md
+++ b/README.md
@@ -892,6 +892,10 @@ docker run -d --name cloak -p 127.0.0.1:9222:9222 cloakhq/cloakbrowser \
 # Headed mode (renders to Xvfb inside container)
 docker run -d --name cloak -p 127.0.0.1:9222:9222 cloakhq/cloakbrowser \
   cloakserve --headless=false
+
+# Reap disconnected per-seed browser processes after 5 minutes
+docker run -d --name cloak -p 127.0.0.1:9222:9222 cloakhq/cloakbrowser \
+  cloakserve --idle-timeout=300
 ```
 
 Stop the server:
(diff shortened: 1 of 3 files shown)


Cloak-HQ:
Thanks for this, @Kumario1 — clean implementation and well-tested. Merged. 🎉

I verified it live in Docker against the real binary before merging:

- **Reaping**: 3 unique seeds spun up 3 Chrome processes (~2.5GB RSS). After all clients disconnected, the reaper killed every process and removed the profile dirs once `--idle-timeout` elapsed — memory dropped back to zero. Exactly what #346 needed.
- **Reconnect safety**: reconnecting to a seed before the timer fires cancels the pending cleanup and keeps the same process alive — no risk of reaping an active client.
- **Backward compatible**: with no `--idle-timeout` set, nothing is reaped (default `0`), so existing setups are unaffected.

Nice touch exposing `idle_timeout` and `idle_cleanup_pending` in `GET /` — made it easy to observe the state machine.

This ships in the next release. Let us know if it resolves the OOM growth on your end!

PR #362, opened 2026-06-09
security: fail-closed binary checksum verification + version-tag validation

rafaelfiguereod-stack:
## Summary

Two small, self-contained, backward-compatible hardening changes to the binary supply chain (Python + TS clients), with new tests.

### 1. Checksum verification fails closed (was fail-open)
`_verify_download_checksum` / `verifyDownloadChecksum` previously **logged a warning and used the binary** when `SHA256SUMS` could not be fetched, or had no entry for the platform archive. Since every release publishes `SHA256SUMS` (and there's a build-provenance attestation), the client should enforce it:
- Missing checksums / missing entry now **raise** instead of proceeding.
- The existing `CLOAKBROWSER_SKIP_CHECKSUM=true` escape hatch is preserved (custom mirrors / offline builds).
- `_fetch_checksums` now skips an empty parse (e.g. a redirect to an HTML 404 page) and falls through to the next mirror, so a misbehaving primary can't shadow a good fallback or satisfy the check with junk. `_parse_checksums` requires a valid 64‑hex digest per line (the TS regex already did).

I verified that the live release `SHA256SUMS` is reachable and contains each platform's entry, so this does **not** break installs.

### 2. Version‑tag validation
The Chromium version comes from the GitHub release **tag name** and flows into cache paths, download URLs, and (Windows, TS client) the PowerShell zip‑extraction command. Added `is_safe_version_tag` / `isSafeVersionTag` (`^\d+(\.\d+){0,9}$`):
- `_get_latest_chromium_version` / `getLatestChromiumVersion` skip any release whose version isn't plain dotted‑numeric.
- Cache‑marker versions are validated before use in `get_effective_version`.
- TS zip extraction passes the archive/destination paths to PowerShell via **environment variables** instead of interpolating them into the `-Command` string, so a path/version containing a quote can't be parsed as code.

## Tests
- New `tests/test_supplychain_hardening.py` (fail‑closed, mirror fall‑through, version validation) — passing.
- `tsc --noEmit` clean. (Pre‑existing Windows‑only failures in `test_update`/`test_config`/`test_extract` are unrelated — they hard‑code `.tar.gz`/POSIX paths.)

## Deliberately NOT included here (to avoid duplicating existing open PRs)
- Opt‑in `cloakserve` authentication — already proposed in #248.
- Archive‑extraction path‑ancestry hardening — already proposed in #227 (Python side); the TS `extractTar` guard could get the same treatment as a follow‑up.

Happy to adjust scope, split, or rename anything if preferred.

### DIFF
--- a/cloakbrowser/config.py
+++ b/cloakbrowser/config.py
@@ -5,9 +5,20 @@ from __future__ import annotations
 import os
 import platform
 import random
+import re
 from pathlib import Path
 
 from ._version import __version__
+
+# A Chromium version tag is a plain dotted-numeric string (e.g. "146.0.7680.177.5").
+# Validate any externally sourced version (GitHub release tags, cache markers)
+# before it flows into cache paths or download URLs.
+_SAFE_VERSION_RE = re.compile(r"^[0-9]+(\.[0-9]+){0,9}$")
+
+
+def is_safe_version_tag(version: str) -> bool:
+    """Return True if *version* is a safe dotted-numeric version string."""
+    return isinstance(version, str) and bool(_SAFE_VERSION_RE.match(version))
 
 # ---------------------------------------------------------------------------
 # Chromium version shipped with this release.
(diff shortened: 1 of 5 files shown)


rafaelfiguereod-stack:
Closing this for now to keep your review queue tidy — it's been open a while and I don't want it lingering as noise. The work is still on my branch, so I'm happy to revive and rebase it if this is of interest down the line. Thanks for the project either way.

PR #385, opened 2026-06-22
Add first-party .NET 8 / C# client

evelaa123:
## First-party .NET 8 / C# client

As discussed in the issue, this adds the .NET client in-repo under `dotnet/`, following the conventions you laid out. It's a complete, faithful port of the CloakBrowser wrapper built on `Microsoft.Playwright`, mirroring the Python and JS clients module-for-module.

### Conventions followed

- **Layout:** `dotnet/` at the repo root, alongside `cloakbrowser/` (Python) and `js/`. Self-contained — its own `CloakBrowser.sln`, `src/` (library + CLI + Roslyn generator), `tests/`, `examples/`. Nothing outside `dotnet/` is touched.
- **Namespacing:** top-level `CloakBrowser` namespace mirroring the Python/JS public surface.
- **Versioning:** tracks the wrapper version — this is `0.4.0`, matching upstream. (Happy to let you drive version sync across the three wrappers.)
- **CI / publishing / main README:** untouched — those are yours. The only readme here is `dotnet/README.md`, which feeds the NuGet package page; structure and content are a proposal, polish away.

### What's ported

- **Binary lifecycle:** download / cache / auto-update with the pinned Ed25519 signature verification (`SHA256SUMS.sig`, anti-downgrade), archive extraction with a zip-slip guard.
- **Launch:** stealth args, proxy (HTTP/HTTPS + SOCKS5 with inline credentials), GeoIP timezone/locale from the proxy exit IP (MaxMind GeoLite2), WebRTC IP spoofing, Widevine seeding, persistent contexts, and real window geometry on headed launches (no fixed emulated viewport over the real window) applied on every page/context path.
- **CloakBrowser Pro (`license_key`):** a `LicenseKey` launch option resolved from param → `CLOAKBROWSER_LICENSE_KEY` → `~/.cloakbrowser/license.key`, with 24h cached validation, the Pro binary gated by the same pinned signature as the free binary, and a clear error rather than a silent downgrade when a valid key's Pro download/signature check fails. Mirrors `validate_license` / `LicenseInfo` / `tier`.
- **Transparent `Humanize = true`:** Bézier mouse motion, human typing, scrolling, and actionability checks — standard Playwright call sites are humanized with no code changes, exactly like the Python/JS `humanize=True` flag.
- **CLI:** `install` / `info` / `update` / `clear-cache`.

### Note on the source generator (per your request)

.NET's Playwright exposes its surface as **sealed interfaces** implemented by internal sealed classes, so the runtime monkey-patching the Python/JS clients use to make humanize transparent isn't available. The .NET client uses a **decorator** instead: each Playwright object is wrapped in a class that implements the same interface, intercepts only the handful of interaction methods, and forwards everything else verbatim.

Hand-writing the "forward to `_inner`" boilerplate for hundreds of members across 8 interfaces would be unmaintainable, so a Roslyn `IIncrementalGenerator` emits it at compile time: annotate a partial decorator with `[GenerateInterfaceDelegation(typeof(IPage))]`, and the generator emits a public forwarding member for every interface member the class doesn't already declare.

It's fully statically typed with **no reflection on the hot path** (`DispatchProxy` is used only in test fakes), and it faithfully reproduces async signatures, generics + constraints, `ref`/`out`/`in`/`params`, optional/default values, properties, indexers and events. A signature-based "already declared" check is what leaves the hand-written humanized methods untouched.

> Full write-up: [`dotnet/DESIGN-SOURCE-GENERATOR.md`](dotnet/DESIGN-SOURCE-GENERATOR.md).

### Tests

**242 total (241 passing / 1 skipped)**, all browser-free (production wrappers exercised through DispatchProxy-backed Playwright fakes; `InternalsVisibleTo` for internal seams). The single skipped test is an end-to-end click-timing check that requires a real browser — verified manually instead.

Coverage includes version/checksum parsing, zip-slip, Ed25519 signature (valid / tampered / bad-sig / downgrade / missing-entry / transient-fetch, free + Pro), license key resolution + cache TTL/expiry, Bézier math, non-ASCII keyboard routing, pointer-events fail-open, the shared timeout budget, focus-aware press/clear, the transparent-wrapper delegate-vs-humanize behavior, and config merge immutability. License/signature tests are ported from `tests/test_license.py`, `js/tests/license.test.ts`, and the signature suites. Build is clean: **0 warnings**.

I'll maintain the .NET client going forward, same as the humanize layer. Anything here you'd like adjusted, just say and I'll turn it around.

Looking forward to your review — happy to adjust, re-test, or rework anything you'd like before merge. Just point me at it.

### Build & test locally

```bash
cd dotnet
dotnet build CloakBrowser.sln -c Release   # 0 warnings
dotnet test  CloakBrowser.sln -c Release   # 242 total (241 passed / 1 skipped)


Cloak-HQ:
Thanks for this — it's a genuinely strong port. We diff'd it against the Python reference across the board and the things that matter are all correct: pinned signature key, fail-closed verification, stealth args, version table, download/anti-downgrade/Pro routing, and the humanize timings all match. The decorator + source-generator approach is clean, and the design note made the generator easy to review. A few changes before we merge:

**1. A handful of interactions skip humanization on `Frame` and `ElementHandle`.** Because they aren't hand-written, the generator forwards them raw to the inner Playwright object:
- `Frame.SelectOptionAsync` (all 6 overloads) and `Frame.DragAndDropAsync`
- `ElementHandle.SelectOptionAsync` and `SetCheckedAsync`

The Python client humanizes all of these, and iframes are exactly where it matters. Could you hand-write the missing overrides? It'd also be great to extend the "no interaction method is generator-delegated" guard test (currently Locator-only) to Page/Frame/ElementHandle/Mouse/Keyboard, so any future fall-through fails the build.

**2. `Download.cs:994` — `double.Parse` is missing `InvariantCulture`.** The siblings at `:1119`/`:1166` have it. On comma-decimal locales (de-DE, fr-FR) the update-check timestamp misparses, so the rate-limit breaks. One-liner.

**3. `HumanScroll.cs:83` throws on headed launches** (`no_viewport` → `ViewportSize` is null). We just landed a fix on main (`9c3ed2d`) that falls back to live `window.innerWidth/innerHeight` — please pull that into the scroll path when you rebase.

**Minor / optional polish:** lowercase the proxy hostname to match Python's output; the `valid` license field and base64-sig whitespace handling are slightly stricter than Python (fine as-is).

**On the design note:** thanks for writing it — it did its job for the review. No need to carry it in the merged tree; we'll keep it internally, so go ahead and drop `DESIGN-SOURCE-GENERATOR.md` and the two README links to it. The other README tweaks are on us post-merge.

Really nice work overall. Ping us once these are in and we'll get it merged.

evelaa123:
Thanks for the thorough review — really appreciate the line-by-line diff against the Python reference. All four items are in:

**1. Humanization on Frame / ElementHandle.** Hand-wrote the missing overrides so nothing falls through to the raw inner object:
- `Frame.SelectOptionAsync` (all 6 overloads) + `Frame.DragAndDropAsync`
- `ElementHandle.SelectOptionAsync` (all 6 overloads) + `SetCheckedAsync`

These now run the same pre-roll path as the Locator versions (mirrors the Python client). I also extended the "no interaction method is generator-delegated" guard test from Locator-only to cover **Page / Frame / ElementHandle / Mouse / Keyboard**, so any future fall-through fails the build.

**2. Download.cs InvariantCulture.** Fixed the `double.Parse` that was missing `CultureInfo.InvariantCulture` — it now matches its siblings, so the update-check timestamp parses correctly on comma-decimal locales (de-DE, fr-FR) and the rate-limit no longer breaks.

**3. HumanScroll null viewport.** Pulled the fix into the scroll path on rebase — falls back to live `window.innerWidth/innerHeight` when `ViewportSize` is null, so headed launches with no_viewport no longer throw.

**Optional polish:** also lowercased the proxy hostname to match Python's output. Left the license-field / base64-sig whitespace handling as-is per your note.

**Design note:** dropped `DESIGN-SOURCE-GENERATOR.md` and the two README links to it, as requested.

Build is clean (0 warnings); test count went 242 → 307 (306 passed, 1 browser-only skip). All in and rebased — ready when you are.
Also squashed the branch into a single commit for a clean merge

Cloak-HQ:
Merged — thanks for the fast, thorough turnaround. We verified everything end-to-end on our side: all four items in, the humanized Frame/ElementHandle overrides run the same pre-roll path as the Locator versions, the guard test now covers every interactive wrapper (so a future fall-through fails the build), `InvariantCulture` and the scroll fallback are correct, and the suite is green (306 passed / 1 browser-skip, 0 warnings).

You're now the maintainer of the .NET client, same as the humanize layer. We'll handle the rest of the plumbing on our side — wiring the README into the NuGet package page, adding the `dotnet test` gate to CI, and publishing to NuGet under the CloakHQ account. We may also do a small pass on `dotnet/README.md` (Pro section + a couple of usage notes); structure stays yours.

Appreciate the contribution — let us know if anything comes up as we get it onto NuGet.

PR #429, opened 2026-07-08
fix: humanize=True resolves iframe locators in the owning frame (#428)

### LINKED ISSUE #428: [Bug] Locator click/fill/etc inside subframes (iframes) fails with ElementNotAttachedError when humanize=True
## Description
When `humanize=True` is enabled, all `Locator` interaction methods (such as `click()`, `fill()`, `hover()`, etc.) fail with an `ElementNotAttachedError` (or `TimeoutError`) if the element is located inside a subframe (iframe).

This occurs because `cloakbrowser`'s class-level patching on `Locator` redirects the interaction to the main `Page` object (e.g. `self.page.click(selector)`), which evaluates the selector on the top-level document rather than within the iframe context.

## Steps to Reproduce

Create a simple page with an iframe and attempt to click a button inside it:

```python
from playwright.sync_api import sync_playwright
from cloakbrowser import launch_context

with sync_playwright() as p:
    # 1. Start context with humanize=True
    context = launch_context(headless=True, humanize=True)
    page = context.new_page()

    # 2. Set page content with an iframe containing a button
    html_content = "<html><body><iframe name='myframe' srcdoc='<html><body><button id=\"btn\">Click Me</button></body></html>'></iframe></body></html>"
    page.goto(f"data:text/html,{html_content}")
    page.wait_for_load_state("domcontentloaded")

    # 3. Locate the button inside the iframe
    frame = page.frame(name="myframe")
    button_locator = frame.locator("#btn")

    # 4. Attempt to click
    print("Clicking button in iframe...")
    try:
        button_locator.click(timeout=3000)
        print("Success!")
    except Exception as e:
        print("Failed with exception:")
        import traceback
        traceback.print_exc()

    context.close()
```

### Expected Behavior
The button inside the iframe should be clicked successfully.

### Actual Behavior
The script raises a `cloakbrowser.human.actionability.ElementNotAttachedError` because the click is routed to the main page where the selector `#btn` does not exist:

```text
playwright._impl._errors.TimeoutError: Locator.wait_for: Timeout 2000ms exceeded.
...
cloakbrowser.human.actionability.ElementNotAttachedError: Element "#btn" failed attached check: element not found in DOM
```

---

## Root Cause Analysis

In `cloakbrowser/human/__init__.py`, inside `_patch_locator_class_sync()`:
```python
def _is_humanized(self):
    return hasattr(self.page, '_original')

def _humanized_click(self, **kwargs):
    if _is_humanized(self):
        self.page.click(_get_selector(self), **_forward_kwargs(kwargs))
    else:
        _orig_click(self, **kwargs)
```

1. Even if a locator is created from a subframe (e.g. `frame.locator("#btn")`), its `self.page` property still returns the main `Page` object. Since the main page has `_original` attached, `_is_humanized(self)` evaluates to `True`.
2. As a result, `_humanized_click` delegates the click to `self.page.click(...)`.
3. `self.page.click` runs the check on the main page DOM instead of the frame DOM, failing to find the selector.

The same issue affects all patched locator methods: `fill`, `type`, `hover`, `dblclick`, `check`, `uncheck`, `set_checked`, `select_option`, `press`, `press_sequentially`, and `clear`.

---

## Proposed Fix / Workaround

Before routing the call to the humanized page method, check if the locator belongs to a subframe. If it does, bypass the humanized layer and fallback to the native Playwright method (which supports frame contexts natively), or route it through the correct frame instance:

```python
def create_wrapped_method(name, orig_method):
    def wrapped(self, *args, **kwargs):
        # Check if the locator's frame is different from the main frame
        impl_frame = getattr(self._impl_obj, '_frame', None)
        page_impl = getattr(self.page, '_impl_obj', None)
        main_frame = getattr(page_impl, 'main_frame', None) if page_impl else None
        
        is_subframe = impl_frame and main_frame and (impl_frame != main_frame)
        
        if is_subframe:
            # Temporarily hide '_original' to force fallback to native Playwright click
            page = self.page
            original = getattr(page, '_original', None)
            has_orig = hasattr(page, '_original')
            if has_orig:
                try:
                    delattr(page, '_original')
                except AttributeError:
                    pass
            try:
                return orig_method(self, *args, **kwargs)
            finally:
                if has_orig:
                    setattr(page, '_original', original)
        else:
            return orig_method(self, *args, **kwargs)
    return wrapped
```

Cloak-HQ:
Fixes #428.

## Problem

With `humanize=True`, any `Locator` action on an element **inside an iframe** (`frame.locator("#btn").click()`, `fill()`, `hover()`, …) failed with `cloakbrowser.human.actionability.ElementNotAttachedError: element not found in DOM`. Native Playwright (`humanize=False`) clicked the same element fine.

## Root cause

The humanize layer resolved every selector against the **main frame**, never the owning sub-frame. Two independent misroutes in `cloakbrowser/human/__init__.py`:

1. **Locator wrapper** — `_humanized_click` etc. called `self.page.click(self._impl_obj._selector, …)`. A frame locator's `.page` is the *main* Page, so `#btn` was searched on the top document.
2. **Frame wrapper** — `_frame_click(sel)` → `page.click(sel)`, discarding the frame identity.

The mouse half was already cross-frame-correct (`bounding_box()` on a frame locator returns page/viewport-space coords), so only *element resolution* was wrong.

## Fix (keeps humanization inside iframes)

Mirrors the design already shipping in the JS wrapper:

- **Frame wrapper** now resolves through `frame.locator(selector)`, does a native `scroll_into_view_if_needed`, reads the page-space `bounding_box`, detects input type in the frame's own context, then runs the **humanized** Bézier mouse-move + human click on the page-level mouse (native-frame fallback if the box can't be read).
- **Locator wrapper** routes a sub-frame locator to its owning frame's now-humanized method; main-frame locators are unchanged (still go through the Page — byte-for-byte the same path). Unknown/detached frames fall back to native Playwright (frame-correct).

Sync **and** async. No changes to the page-level primitives, so the main-frame path carries zero regression risk.

The JS and .NET wrappers were already frame-correct — this is a **Python-only** change (verified by running a real same-origin iframe through all three wrappers).

## Testing

- **New unit tests** (mocked, no binary): sub-frame Locator routing (sub-frame → owning frame, main-frame → page, unpatched → native), frame-humanized resolution + native fallback, sync + async.
- **New integration tests** (`@pytest.mark.slow`, real binary, local same-origin iframe over HTTP): humanized click + fill inside an iframe, async click, and a `humanize=False` parity control.
- Existing humanize unit suite: **71 passed** (no main-frame regression).
- On the v148 Pro binary in Docker: the 4 new iframe integration tests pass; pre-existing humanize browser tests pass; matrix **group 9 (Humanize)** and groups **1/2/3/5** = **57/57 passed, 0 failed**.

## Ask

@evelaa123 — as the humanize layer's main author, could you review this and re-run the humanize tests on your side (especially the new iframe integration tests, and any humanize flows you rely on)? Would like a second pair of eyes on the frame-routing before it merges.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Changes are tagged: **[wrapper]** for Python/JS wrapper, **[binary]** for Chromi
 ## [Unreleased]
 
 - **[wrapper]** Fix JS CLI silently exiting with no output when run via `npx`/`node_modules/.bin` (#427). The entry-point guard compared `import.meta.url` to an unresolved `process.argv[1]`; symlinked bin installs (npm/pnpm/npx) never matched, so no subcommand ran. Now realpath-resolves the invoked path before comparing.
+- **[wrapper]** Fix `humanize=True` breaking interactions with elements **inside iframes** (#428). `frame.locator("#btn").click()` / `fill()` / `hover()` / etc. (and the frame-level `frame.click(...)` equivalents) now resolve the selector in the owning sub-frame's own document instead of misrouting to the top page, which previously raised `ElementNotAttachedError`. Humanized mouse motion is preserved inside iframes. Python only (the JS and .NET wrappers were already frame-correct).
 
 ## [0.4.9] — 2026-07-09
 
--- a/cloakbrowser/human/__init__.py
+++ b/cloakbrowser/human/__init__.py
@@ -380,6 +380,30 @@ def _patch_locator_class_sync():
     def _get_cfg(self):
         return getattr(self.page, '_human_cfg', None)
 
+    def _route_target(self):
+        """Resolution target for a humanized locator action (#428).
+
+        Main-frame locators return the Page (unchanged behavior). A locator that
+        belongs to a *sub-frame* returns that frame's humanized wrapper so the
+        selector resolves in the frame's own document. Returns ``None`` when the
+        owning sub-frame is known but not yet humanized, so the caller falls back
+        to the native Playwright locator method (frame-correct, un-humanized).
+        """
+        page = self.page
+        impl_frame = getattr(getattr(self, "_impl_obj", None), "_frame", None)
+        if impl_frame is None:
+            return page
+        try:
+            if impl_frame is page.main_frame._impl_obj:
+                return page
+            for f in page.frames:
+                if getattr(f, "_impl_obj", None) is impl_frame:
+                    return f if getattr(f, "_human_patched", False) else None
+            # Frame not among page.frames (unknown/detached) -> legacy Page path.
+            return page
+        except Exception:
+            return page
+
     def _forward_kwargs(kwargs):
         out = {}
         if "timeout" in kwargs:
@@ -392,31 +416,51 @@ def _patch_locator_class_sync():
 
     def _humanized_fill(self, value, **kwargs):
         if _is_humanized(self):
-            self.page.fill(_get_selector(self), value, **_forward_kwargs(kwargs))
+            tgt = _route_target(self)
+            if tgt is None:
+                _orig_fill(self, value, **kwargs)
+            else:
+                tgt.fill(_get_selector(self), value, **_forward_kwargs(kwargs))
         else:
             _orig_fill(self, value, **kwargs)
 
     def _humanized_click(self, **kwargs):
         if _is_humanized(self):
-            self.page.click(_get_selector(self), **_forward_kwargs(kwargs))
+            tgt = _route_target(self)
+            if tgt is None:
+                _orig_click(self, **kwargs)
+            else:
+                tgt.click(_get_selector(self), **_forward_kwargs(kwargs))
         else:
             _orig_click(self, **kwargs)
 
     def _humanized_type(self, text, **kwargs):
         if _is_humanized(self):
-            self.page.type(_get_selector(self), text, **_forward_kwargs(kwargs))
+            tgt = _route_target(self)
+            if tgt is None:
+                _orig_type(self, text, **kwargs)
+            else:
+                tgt.type(_get_selector(self), text, **_forward_kwargs(kwargs))
         else:
             _orig_type(self, text, **kwargs)
 
     def _humanized_dblclick(self, **kwargs):
         if _is_humanized(self):
-            self.page.dblclick(_get_selector(self), **_forward_kwargs(kwargs))
+            tgt = _route_target(self)
+            if tgt is None:
+                _orig_dblclick(self, **kwargs)
+            else:
+                tgt.dblclick(_get_selector(self), **_forward_kwargs(kwargs))
         else:
             _orig_dblclick(self, **kwargs)
 
     def _humanized_hover(self, **kwargs):
         if _is_humanized(self):
-            self.page.hover(_get_selector(self), **_forward_kwargs(kwargs))
+            tgt = _route_target(self)
+            if tgt is None:
+                _orig_hover(self, **kwargs)
+            else:
+                tgt.hover(_get_selector(self), **_forward_kwargs(kwargs))
         else:
             _orig_hover(self, **kwargs)
 
@@ -450,6 +494,10 @@ def _patch_locator_class_sync():
 
     def _humanized_check(self, **kwargs):
         if _is_humanized(self):
+            tgt = _route_target(self)
+            if tgt is None:
+                _orig_check(self, **kwargs)
+                return
             fwd = _forward_kwargs(kwargs)
             cfg = _get_cfg(self)
             if cfg and cfg.idle_between_actions:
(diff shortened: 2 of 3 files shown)


evelaa123:
Reviewed and tested locally on Windows

**What I ran:**
- Routing unit tests (`TestLocatorSubframeRouting`, `TestLocatorSubframeRoutingAsync`, `TestFrameHumanizedRouting`) — 6/6 pass. Sub-frame → owning frame, main-frame → page, unpatched → native fallback all confirmed.
- Full non-browser humanize unit suite — green, no main-frame regression.
- The 4 new `@pytest.mark.slow` iframe integration tests on a real binary — 4/4 pass (sync click + fill, async click, `humanize=False` parity control).
- Manual end-to-end with a local same-origin iframe: humanized `frame.locator("#btn").click()` → `CLICKED`, `frame.locator("#inp").fill("hello")` → `hello`. Confirmed the cursor tracks into the frame in page-space coords instead of misrouting to the top document.

Also verified the CDP holder path: `patch_page_async` sets `page._cdp_session_holder` before frame patching, so the async frame path reuses the same session. No issue there.

**Suggested extra tests (optional, but worth it):** the mocked routing tests mock out `human_click`, so they wouldn't catch an argument-order regression. Two small unit tests would close that gap, built on the existing `_make_frame_for_routing` helper:

```python
class TestFrameClickArgContract:
    """#428 guard: _frame_click must call human_click(raw, is_input, cfg) in order.
    The routing tests mock human_click, so a swap would slip through."""

    def test_frame_click_passes_is_input_bool_and_cfg(self):
        from unittest.mock import patch as mock_patch
        from cloakbrowser.human.config import HumanConfig
        frame, page, _ = _make_frame_for_routing(is_input=False)  # button
        with mock_patch("cloakbrowser.human.human_move"), \
             mock_patch("cloakbrowser.human.human_click") as clk:
            frame.click("#btn")
        args = clk.call_args[0]
        assert isinstance(args[1], bool) and args[1] is False   # is_input
        assert isinstance(args[2], HumanConfig)                 # cfg

    def test_frame_click_marks_is_input_true_for_input(self):
        from unittest.mock import patch as mock_patch
        frame, page, _ = _make_frame_for_routing(is_input=True)
        with mock_patch("cloakbrowser.human.human_move"), \
             mock_patch("cloakbrowser.human.human_click") as clk:
            frame.click("#inp")
        assert clk.call_args[0][1] is True


class TestFrameFillFocus:
    def test_frame_fill_clicks_before_typing(self):
        from unittest.mock import patch as mock_patch
        frame, page, _ = _make_frame_for_routing(is_input=True)
        with mock_patch("cloakbrowser.human.human_move"), \
             mock_patch("cloakbrowser.human.human_click"), \
             mock_patch("cloakbrowser.human.human_type") as htype:
            frame.fill("#inp", "hello")
        assert htype.called and htype.call_args[0][2] == "hello"

    def test_frame_fill_falls_back_when_type_raises(self):
        from unittest.mock import patch as mock_patch
        frame, page, _ = _make_frame_for_routing(is_input=True)
        with mock_patch("cloakbrowser.human.human_move"), \
             mock_patch("cloakbrowser.human.human_click"), \
             mock_patch("cloakbrowser.human.human_type", side_effect=RuntimeError("boom")):
            frame.fill("#inp", "hello")  # must not raise — native fallback
```

All four pass locally against the current diff. Not required for merge — just pins the `human_click` arg contract and the `_frame_fill` fallback so a future refactor of the `_move_to_frame_selector` return tuple can't silently break the call order.

**One non-blocking note (follow-up ticket, not this PR):** `_route_target` does a linear `for f in page.frames` scan on every humanized locator action. Fine for typical pages, but on pages with hundreds of frames (ad/tracker iframes) it's O(n) per click/fill. Worth caching the `impl_frame → wrapper` mapping at some point. Purely a perf nit — correctness is solid.

Cloak-HQ:
Thanks for the thorough review and for re-running everything on Windows — the routing / integration / manual pass all matching gives good confidence here.

Added your two suggested test guards before merge:
- `TestFrameClickArgContract` — pins `human_click(raw, is_input, cfg)` argument order (the routing tests mock `human_click`, so a swap would've slipped through, exactly as you noted).
- `TestFrameFillFocus` — pins that `_frame_fill` clicks before typing and falls back to native when `human_type` raises.

All 4 pass against the merged diff, and the existing humanize unit suite stays green.

On the `_route_target` linear `for f in page.frames` scan — agreed it's worth addressing, and you're right that it's a separate follow-up. One subtlety for whoever picks it up: `_route_target` has three outcomes (main-frame → page, known-patched sub-frame → wrapper, known-**un**patched sub-frame → native fallback), so a cache that only records patched frames would let an unpatched sub-frame miss and fall back to the main frame — reintroducing #428. A correct cache needs to track frame lifecycle (create / navigate / detach). Filing it as a tracking issue.

Merged as `1ae7605`. Thanks again.

PR #454, opened 2026-07-19
fix(cli): run main() when invoked via a symlinked bin (npx) (#427)

### LINKED ISSUE #427: CLI silently exits with no output when run via npx
**Description:**

`npx cloakbrowser info` (and every other CLI subcommand) produces no output and exits 0. I suspect the CLI silently skips `main()` because of something with the entry-point guard at the bottom of `dist/cli.js`:

```js
const invokedPath = process.argv[1];
if (invokedPath && import.meta.url === pathToFileURL(invokedPath).href) {
    main();
}
```

**CloakBrowser version:** 0.4.8 (also present in 0.4.7)

**Wrapper:** JavaScript

**Environment:** macOS (Darwin 24.6.0), arm64, no Docker. Node v25.8.2, pnpm 11.10.0 (also `npx`).

**Launch options:** N/A (CLI issue, not a browser launch issue)

**Tested with a different IP or proxy?** N/A

**Works outside Docker / on host machine?** Not using Docker (bug occurs on host)

**Steps to reproduce:**

1. `npm install cloakbrowser` (or pnpm)
2. `npx cloakbrowser info`

**Error output / screenshots:**

```sh
$ npx cloakbrowser info
$ echo $?
0
```

**Dockerfile (if applicable):** N/A

dsitmilis:
## Summary
Fixes #427 — `npx cloakbrowser info` (and every subcommand) produced no output and exited 0. The CLI entry guard compared `import.meta.url` against `pathToFileURL(invokedPath)` but only realpathed the invoked side; under npx/pnpm-exec the bin is a symlink (`node_modules/.bin/cloakbrowser` -> `dist/cli.js`) and the two realpaths did not match, so `main()` never ran.

## Fix
- js/src/cli.ts: extract `isCliEntry(invokedPath, metaUrl)` which realpaths **both** sides before comparing and returns `false` (never auto-running `main()`) on any resolution error. The module-level guard now uses it.
- js/tests/cli.test.ts: unit tests for `isCliEntry` — symlink resolution, missing path, unrelated file, and undefined `argv[1]`.

## Validation
- `node_modules/.bin/vitest run tests/cli.test.ts` -> 6 passed.
- End-to-end: a symlinked bin shim invoking `dist/cli.js` now prints `CloakBrowser diagnostics` (exit 0) instead of silently exiting. Importing `cli.js` in tests does NOT auto-run `main()` (the import-by-test case stays safe).
- `tsc` build clean.

Closes #427.

### DIFF
--- a/js/src/cli.ts
+++ b/js/src/cli.ts
@@ -25,7 +25,7 @@ import { countFontsPresent, WINDOWS_FONT_TELLS, OFFICE_FONT_TELLS } from "./font
 import { resolveLicenseKey, validateLicense, getProLatestVersion, getActiveSessionCount, type LicenseInfo } from "./license.js";
 import { execFileSync } from "node:child_process";
 import { createRequire } from "node:module";
-import { pathToFileURL } from "node:url";
+import { fileURLToPath, pathToFileURL } from "node:url";
 import fs from "node:fs";
 import os from "node:os";
 import path from "node:path";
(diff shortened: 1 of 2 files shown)


Cloak-HQ:
Thanks for the contribution and for adding focused tests.

We reproduced the original failure on `0.4.8`: `npx cloakbrowser info` exited `0` with no output. However, this was already fixed by commit `045219b` and shipped in `0.4.10`.

We retested `0.4.10`, the latest release (`0.5.6`), and current `main` through an npm-style symlink; all correctly run the CLI and print diagnostics. Node resolves `import.meta.url` to the real module path, while the existing guard realpaths the invoked path, so resolving both sides is not required.

Since the reported issue is already resolved, we’ll close this PR as superseded. Thanks again for taking the time to investigate it.

PR #470, opened 2026-07-27
fix: preserve HTTP proxy credentials during GeoIP resolution

### LINKED ISSUE #469: Preserve HTTP proxy credentials during GeoIP resolution
When `geoip` is enabled and an authenticated HTTP proxy is provided with separate `username` and `password` fields, `extractProxyUrl()` returns only `proxy.server`. Unlike SOCKS proxies, the HTTP proxy URL is not reconstructed with its credentials.

`maybeResolveGeoip()` then uses this unauthenticated URL for GeoIP resolution. Because `Proxy-Authorization` is only added when credentials are present in the URL, the proxy rejects the CONNECT request with `407 Proxy Authentication Required`.

```javascript
{
  geoip: true,
  proxy: {
    server: "http://<proxy-host>:<proxy-port>",
    username: "<username>",
    password: "<password>"
  }
}
```

Relevant code:

- [`extractProxyUrl()` only reconstructs credentials for SOCKS proxies](https://github.com/CloakHQ/CloakBrowser/blob/093a6646016f7068ff1bcf92a7ec4fd226f694b0/js/src/geoip.ts#L414-L422)
- [`maybeResolveGeoip()` passes the extracted URL to GeoIP resolution](https://github.com/CloakHQ/CloakBrowser/blob/093a6646016f7068ff1bcf92a7ec4fd226f694b0/js/src/geoip.ts#L464)
- [`Proxy-Authorization` requires credentials in the proxy URL](https://github.com/CloakHQ/CloakBrowser/blob/093a6646016f7068ff1bcf92a7ec4fd226f694b0/js/src/geoip.ts#L262-L275)

Expected behavior: authenticated HTTP proxy objects should be reconstructed with their credentials before GeoIP resolution, just like authenticated SOCKS proxy objects.

> CloakBrowser version: `150.0.7871.114.3`  
> Wrapper: JavaScript (`cloakbrowser@0.4.11`)

ishiko732:
## Summary

- Preserve HTTP proxy credentials during GeoIP and WebRTC exit-IP resolution.
- Reuse existing credential-aware proxy URL reconstruction in the JavaScript, Python, and .NET SDKs.
- Keep unauthenticated HTTP proxy behavior unchanged.
- Add regression tests, including a JavaScript CONNECT proxy authentication test.

## Testing

- JavaScript: `npm test -- tests/geoip.test.ts` — 17 passed
- JavaScript: `npm run typecheck` — passed
- Python: `uv run --extra dev --extra geoip pytest -q tests/test_proxy.py::TestMaybeResolveGeoip tests/test_geoip.py` — 40 passed
- .NET: `dotnet test dotnet/tests/CloakBrowser.Tests/CloakBrowser.Tests.csproj --filter FullyQualifiedName~ProxyResolverTests` — 12 passed

Closes #469

### DIFF
--- a/cloakbrowser/browser.py
+++ b/cloakbrowser/browser.py
@@ -1053,8 +1053,8 @@ def _normalize_socks_string_url(url: str) -> str:
 def _extract_proxy_url(proxy: str | ProxySettings | None) -> str | None:
     """Extract and normalize proxy URL string from proxy param.
 
-    For SOCKS5 dicts with separate username/password fields, reconstructs
-    the full URL with inline credentials so SOCKS5 auth works.
+    For proxy dicts with separate username/password fields, reconstructs
+    the full URL with inline credentials so proxy auth works.
     """
     if proxy is None:
         return None
@@ -1062,8 +1062,11 @@ def _extract_proxy_url(proxy: str | ProxySettings | None) -> str | None:
         server = proxy.get("server", "")
         if not server:
             return None
-        if _is_socks_proxy(proxy):
-            return _reconstruct_socks_url(proxy)
+        if proxy.get("username"):
+            return (
+                _reconstruct_socks_url(proxy) if _is_socks_proxy(proxy)
+                else _reconstruct_http_url(proxy)
+            )
         return _ensure_proxy_scheme(server)
     return _ensure_proxy_scheme(proxy)
 
--- a/dotnet/src/CloakBrowser/ProxyResolver.cs
+++ b/dotnet/src/CloakBrowser/ProxyResolver.cs
@@ -272,7 +272,9 @@ internal static class ProxyResolver
                 return null;
             case ProxySettings ps:
                 if (string.IsNullOrEmpty(ps.Server)) return null;
-                return IsSocksProxy(ps) ? ReconstructSocksUrl(ps) : EnsureProxyScheme(ps.Server);
+                return IsSocksProxy(ps)
+                    ? ReconstructSocksUrl(ps)
+                    : EnsureProxyScheme(ReconstructHttpUrl(ps));
             case string s:
                 return EnsureProxyScheme(s);
             default:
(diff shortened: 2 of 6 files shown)


Cloak-HQ:
Thanks for the fix, and for the thorough test coverage.

Merged as `3739d7b`.

We reproduced this on Linux before merging, against an authenticated proxy that is not IP-whitelisted, with the exit IP pinned to a sticky session so both proxy forms had to resolve to the same address. Confirmed the settings-object form resolved to a different location than the equivalent inline-credential URL, and that both forms resolve identically with your change.

Worth noting for anyone who hits this: it did not fail loudly. The rejected lookup fell back to resolving the proxy gateway's hostname, which returns a real address, so resolution appeared to succeed while reporting the gateway's timezone and locale instead of the actual exit IP. `--fingerprint-webrtc-ip=auto` was silently dropped in the same path.

All three suites pass on your branch: Python 104, JavaScript 513, .NET 397.

We added a changelog entry and credited you in the README contributors list. This ships in the next release.

ishiko732:
> Worth noting for anyone who hits this: it did not fail loudly. The rejected lookup fell back to resolving the proxy gateway's hostname, which returns a real address, so resolution appeared to succeed while reporting the gateway's timezone and locale instead of the actual exit IP. `--fingerprint-webrtc-ip=auto` was silently dropped in the same path.

Yes, because no error was reported, this bug was very well hidden. I initially thought that the geo configuration simply did not correctly apply the locale settings.  
I also noticed that geo=true does not configure Geolocation (--fingerprint-location). Is there a specific reason why fingerprint-location is not set in this case?

Cloak-HQ:
Your read on the permission model is right. Geolocation is not a passive surface. A site gets a position only if it asks and the user grants it, which is why it has not been a priority the way timezone and locale are, since those leak on every page load without asking.

One correction on the specifics though. `--fingerprint-location` was not a working switch we chose not to wire into `geoip`. It was declared on the command line surface but nothing ever consumed it, so it silently did nothing, and it has since been removed from the binary. Setting it from `geoip=True` would have had no effect at any point.

So the honest state is that geolocation spoofing is not implemented, rather than implemented and left unused.

Thanks for pointing at it. The flag is still listed in the README flag table, which we are removing.

If you have a concrete flow where the spoofed coordinates would matter, we would like to hear it. A real use case is what we are missing to justify building it.

Issue #395, opened 2026-06-25
FingerprintJS still reports browserTampering on Linux with minimal Windows font setup (Chromium 148)


maithanhtrung123:
**Description**
On a Linux VPS running CloakBrowser Pro Chromium 148, FingerprintJS Pro Playground still reports **`browserTampering: detected`** despite using the recommended minimal Windows font setup.

**Environment**
* cloakbrowser@0.4.3
* Pro Chromium 148.0.7778.215.2
* Ubuntu Linux (no Docker)
* Node.js v22.23.0
* Running via `launchPersistentContext()`
* Xvfb (`DISPLAY=:99`)

**Launch options:**

Using `launchPersistentContext()`
import { launchPersistentContext } from "cloakbrowser";
const context = await launchPersistentContext({
  userDataDir: "",
  headless: false,
  humanize: true,
  proxy: "http://user:pass@ip:port",
  geoip: true,
  viewport: { width: 1920, height: 1080 },
  deviceScaleFactor: 1,
  args: [
    "--fingerprint=98293",
    "--fingerprint-platform=windows",
    "--fingerprint-windows-font-metrics",
    "--fingerprint-noise=false",
    "--fingerprint-webrtc-ip=auto",
    "--fingerprint-screen-width=1920",
    "--fingerprint-screen-height=1080"
  ]
});
**Windows fonts**
Installed only the recommended six fonts system-wide:
* Calibri
* Franklin Gothic Medium
* Marlett
* MS UI Gothic
* Segoe UI
* Segoe UI Light

**Reality**
With the minimal Windows font setup and `--fingerprint-windows-font-metrics`, FingerprintJS should report:
* `browserTampering: Not detected`
* or at least a significantly lower tampering score.

Fresh profile results:
✅ bot: Not detected
✅ virtualMachine: Not detected
❌ browserTampering: detected
The issue is reproducible across different residential proxies. Persistent context is used (not regular `launch()`).

**Additional notes**
Compared to a full Windows font directory, the minimal 6-font setup noticeably improves detection:
* `virtualMachine` is no longer detected.
* Overall suspect score is lower.
* The only remaining signal is `browserTampering`.

FingerprintJS also reports:
vpn: You are using a VPN 🌐 (OS mismatch)

The browser fingerprint is reported as:
* Operating System: Windows 11
* User-Agent: Windows NT 10.0; Win64; x64

However, the browser is actually running on a Linux using:
--fingerprint-platform=windows
[...]

PR #472, opened 2026-07-27
fix: align Windows font metrics when using fonts dir

### LINKED ISSUE #471: Blocked by Datadome
Description: 

I’ve been using CloakBrowser on my home computer for several months, specifically for accessing https://www.allegro.pl/. At first, there were some problems. I reported these issues in the issue tracker at https://github.com/CloakHQ/CloakBrowser/issues/16. All the problems were eventually resolved, and everything worked fine until Saturday. Since Saturday, every attempt to access [www.allegro.pl](http://www.allegro.pl/) results in a message stating that I’m blocked.
I updated the version from 4.x to 5.2, but nothing changed. I also tried the Pro version using the key from https://cloakbrowser.dev/free/, with the same result. As soon as I open the browser, I’m blocked.

CloakBrowser version:  0.4.x and 0.5.2

Wrapper: Python

Environment:  Debian 13 on VM in Proxmox

Launch options:

```
  browser = await launch_persistent_context_async(
        user_data_dir=browser_profile_path,
        humanize=True,
        human_preset="careful",
        headless=False,
        timezone="Europe/Warsaw",
        geoip=True,
        locale="pl-PL",
        args=[
            "--fingerprint-fonts-dir=/usr/share/fonts/windows",
            "--fingerprint=23029",
            "--fingerprint-platform=windows",
            "--fingerprint-hardware-concurrency=8",
            "--fingerprint-gpu-vendor=NVIDIA Corporation",
            "--fingerprint-gpu-renderer=NVIDIA GeForce RTX 3070",
            "--disable-http2"
        ]
    )

```

Tested with a different IP or proxy? 
My home has internet through a local provider. On another computer in the house with a Windows system, using the same IP address, the internet works just fine.

Error output / screenshots:

Image

expary:
## Summary
- auto-add `--fingerprint-windows-font-metrics` on Linux when the effective persona is Windows and `--fingerprint-fonts-dir` is provided
- keep non-Windows personas and explicit user-supplied metrics flags unchanged
- add Python, JavaScript, and .NET regression coverage

Fixes #471.

## Verification
- `. .venv/bin/activate && pytest tests/test_build_args.py -q`
- `. .venv/bin/activate && pytest tests/ -q -m "not slow"`
- `cd js && npm run build`
- `cd js && npm run typecheck && npm test`
- `dotnet test dotnet/CloakBrowser.sln -c Release --filter BuildArgsTests`
- `dotnet test dotnet/CloakBrowser.sln -c Release --no-build`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Changes are tagged: **[wrapper]** for Python/JS wrapper, **[binary]** for Chromi
 
 ## [Unreleased]
 
+- **[wrapper]** When spoofing Windows on Linux with a user-provided Windows fonts directory, the wrappers now automatically add `--fingerprint-windows-font-metrics` unless you already set it. This keeps font names and text metrics aligned for aggressive font probes such as DataDome while preserving non-Windows personas and explicit user overrides (#471). Python, JavaScript, and .NET.
 - **[wrapper]** **`cloakbrowser info` no longer aborts partway through on a Windows console.** `cmd.exe` defaults to a code page that cannot represent the report's check mark and arrow, so the command stopped with an encoding error at the first one and the remaining diagnostics were never printed. Those marks now degrade to plain text when the console cannot take them, per character. UTF-8 consoles, including Linux, macOS and Windows Terminal, are unchanged. Python.
 - **[wrapper]** **`cloakbrowser info` no longer reports a false launch failure on Windows.** Chromium handles `--version` only on POSIX, so on Windows the switch was ignored and a browser started instead of printing — the probe then timed out and a healthy install was reported as broken, briefly putting a window on screen each run. The check now exits immediately on Windows, without a window, and still fails loudly on a genuinely broken binary; it reports no version there, since nothing is printed. Linux and macOS are unchanged. Python, JavaScript, and .NET.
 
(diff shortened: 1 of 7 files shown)


Cloak-HQ:
Thanks for the PR, and for covering all three wrappers with tests. The intent is right, but we can't take this one.

`--fingerprint-windows-font-metrics` is opt-in on purpose. It doesn't just realign metrics, it changes which font is actually resolved for generic families, so it alters rendered text and canvas output. That's the correct choice for a narrow set of checks, most notably the FingerprintJS tampering model, where we do recommend it and have published a full recipe using it in #395. It's the wrong choice for most other sites, where it changes rendering for no benefit.

Auto-enabling it would flip that behavior on for every Linux user running a Windows persona with a fonts directory, on the basis of a config heuristic rather than the target actually being tested. There are a lot of sites, and this flag is specific to very specific cases.

Worth noting too: the metrics alignment that should apply to every Windows persona already does, automatically, with no flag at all. That part is handled in the binary and leaves every glyph unchanged. This flag is the separate, deliberately narrow case, which is why it's kept behind an explicit opt-in.

On #471 specifically, the fonts directory flag is itself part of what we recommended removing there. The Windows fonts should be installed system-wide rather than pointed at with a flag, and the rest of the profile left to the seed.

The contribution is appreciated. If you hit a case where the flag is clearly needed and isn't being applied, open an issue with the target and we'll take a look.

PR #511, opened 2026-08-19
feat(browser): add launch_chrome_direct() for TCP CDP server mode

WilliamLuuu:
## Problem

Server-style runtimes that want to hand a CDP endpoint to another process (e.g. an MCP browser server) currently route through `launch_persistent_context(user_data_dir, args=["--remote-debugging-port=9222"])`. Playwright's `launch_persistent_context` injects `--remote-debugging-pipe` and owns the connection over that pipe; the TCP `--remote-debugging-port` argument never becomes a usable endpoint for external CDP clients. External tools that `connect_over_cdp` to the advertised port hang or fail, and the orphaned instance wedges the shared profile so every later launch exits immediately with code 0.

## Fix

Adds `launch_chrome_direct()` — a subprocess-based launch that:
- builds the same stealth args via `build_args()` (fingerprints, extensions, locale/timezone),
- launches the patched Chromium directly with `--remote-debugging-address` / `--remote-debugging-port` (TCP, no pipe),
- waits for the CDP listener, then hands back the `Popen` handle for the caller to own,
- **Windows only**: appends `--disable-features=AutoDeElevate`. Since Chrome 131, AutoDeElevate relaunches the browser de-elevated; inside the patched build the re-exec fails and the parent exits 0 instantly — killing the launch. Playwright's default args mask this, which is why it only bites direct launches.

## Verification

- `tests/test_launch.py::test_launch_chrome_direct_exposes_tcp_cdp` — real binary: `/json/version` answers over TCP, no `--remote-debugging-pipe` in argv.
- `tests/test_launch.py::test_launch_chrome_direct_auto_delevate_always_on_windows` — flag always present on Windows, merged (not duplicated) into caller-supplied `--disable-features`, absent on non-Windows.
- End-to-end: Playwright `connect_over_cdp` + `page.goto("https://www.google.com")` works against the direct-launched endpoint.

### DIFF
--- a/cloakbrowser/__init__.py
+++ b/cloakbrowser/__init__.py
@@ -11,7 +11,7 @@ Usage:
     browser.close()
 """
 
-from .browser import launch, launch_async, launch_context, launch_context_async, launch_persistent_context, launch_persistent_context_async, ProxySettings, build_args, maybe_resolve_geoip
+from .browser import launch, launch_async, launch_context, launch_context_async, launch_persistent_context, launch_persistent_context_async, ProxySettings, build_args, maybe_resolve_geoip, launch_chrome_direct
 from .config import CHROMIUM_VERSION, get_default_stealth_args
 from .download import binary_info, check_for_update, clear_cache, ensure_binary
 from .license import CloakBrowserLicenseError, LicenseInfo, validate_license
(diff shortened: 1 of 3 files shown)


WilliamLuuu:
AI did that when i'm sleeping...sry LOL

Issue #512, opened 2026-08-20
CloakBrowser 0.5.8 — False-negative visible actionability check on dropdown menu items


thesob:
# Bug Report: CloakBrowser 0.5.8 — False-negative `visible` actionability check on dropdown menu items
---

## Environment

| | |
|---|---|
| `cloakbrowser` (wrapper) | **0.5.8** (regression) — confirmed working on **0.4.11** |
| Browser binary | v146 (free tier) |
| Playwright | 1.61.0 |
| Python | 3.12.13 |
| OS | Linux 6.8.0-137-generic x86_64 (Docker container, glibc 2.41) |
| Launch mode | `launch_persistent_context(..., headless=False, humanize=True, viewport={&quot;width&quot;: 1920, &quot;height&quot;: 1080})` |

## Summary

After upgrading only the `cloakbrowser` wrapper from `0.4.11` → `0.5.8` (base image + pip package, no other code changes), a previously-reliable `page.click("text=Hoy")` call started failing **100% of the time** with `ElementNotVisibleError`, even though the target element is demonstrably visible and rendered on screen at the exact moment of the check.

## Steps to reproduce (site-specific, described since it's on a private/authenticated site)

Open an open, absolutely-positioned dropdown menu (`.daterangepicker.dropdown-menu, display: block`) containing a list of `<li>` options, e.g.:
```
<div class="daterangepicker dropdown-menu opensright" style="display: block; ...">
  <div class="ranges">
    <ul>
      <li class="active">Hoy</li>
      <li>Últimos 7 Días</li>
      <li>Mes Actual</li>
      ...
    </ul>
  </div>
  <div class="calendar left">...</div>
</div>
```
Call page.click("text=Hoy", timeout=10000) (via the humanized wrapper, humanize=True).
3. Observe `ElementNotVisibleError: Element 'text=Hoy' failed visible check: element is not visible` after the full timeout elapses.
4. Retry with `force=True` → this skips `ensure_actionable()` entirely but still calls `scroll_to_element()`, which now raises `RuntimeError: Element not found while scrolling into view` because `get_box()` returns `None`.

## Expected behavior

The click should succeed (as it reliably did on `0.4.11`), since the element:
- Is attached to the DOM
- Has non-zero size (visually rendered, orange-highlighted `active` state)
- Is within the viewport
- Is not covered by any other element

## Actual behavior

[...]

PR #515, opened 2026-08-21
Fix humanized selector resolution and actionability

Cloak-HQ:
## Summary

- align supported selector resolution and actionability behavior across Python, JavaScript, and .NET
- preserve expected target ordering and validate target identity through humanized actions
- support rendered text and visible descendants when a target has no box of its own
- keep unsupported and complex locator paths on their existing compatibility behavior
- add cross-wrapper unit and real-browser regression coverage

This addresses #512 and incorporates the relevant geometry correction identified in #514 without adding the proposed visibility fallback.

## Validation

- Python focused selector/actionability tests: 112 passed
- Python Preview browser regressions: 12 passed
- JavaScript build and typecheck: passed
- JavaScript tests: 564 passed, 19 skipped
- JavaScript Preview selector regressions: 8 passed
- .NET tests: 453 passed, 1 skipped
- .NET Preview validation: 443 passed, 1 skipped; 10 browser regressions passed
- cross-wrapper resolver parity verified
- `git diff --check` clean


Cloak-HQ:
Merged into main via fast-forward (commits rebased onto latest main; hashes differ so GitHub didn't auto-close). Content is identical to this PR.

